PR #313, opened 2018-09-05
Dapp browser

JamesSmartCell:
Integrate dapp browser.


app/src/main/java/io/stormbird/wallet/web3/entity/Address.java
```diff
@@ -0,0 +1,69 @@
...
+import android.os.Parcelable;
+import android.support.annotation.NonNull;
+import android.text.TextUtils;
+
+import io.stormbird.wallet.util.Hex;
+
+
+public class Address implements Parcelable {
+
+    public static final Address EMPTY = new Address("0000000000000000000000000000000000000000");
```
ligi: not sure if EMPTY is good for 0000..0000 - as this address is often used to burn ether.

colourful-land:
@JamesSmartCell can we remove the reference to contract folder entirely? Since the contract folder will never be updated again. To avoid confusion I will move the contract folder to the test directory as it is the only place which uses it.

app/src/main/java/io/stormbird/wallet/web3/entity/Address.java
JamesSmartCell: Yes, that's true. 

In this case it's an attempt to indicate something is wrong in a theoretical edge condition where a user starts the browser with an invalid wallet. It would be very difficult to get this, but there should be some fallback.

Should it ever be displayed no transactions or signing would be able to be done.

Thankyou for checking out the source!

ligi: You are very welcome - this is the power of open-source - we can learn from each other and increase the quality of the eco-system as a whole!-)
Yea - currently EMPTY is only used for this - but I think it sitting there will lead to it being used for other cases in the future - maybe even detecting it as EMPTY when it should be detected as a normal/recipient/burn address. Just a thought when briefly looking over the code - will have a deeper look after ETHBerlin (btw anyone of your team there?)
Thanks for being libre software - keep up the good work!
PS: one question that also came to my mind when looking over this - what is the reason you inlinded the thing and did not use https://github.com/TrustWallet/Web3View ? As far as I see there is a lot of overlap.

JamesSmartCell: Good point, perhaps it should be named something like ERROR_ADDRESS and scoped to only where it's actually needed.

We're not at ETHBerlin but we might be at Devcon later. We didn't know there was a Web3View in the Trust Repo, otherwise we'd have definitely checked it out. We based our dapp integration on a different open source implementation (Neuron).

PR #315, opened 2018-09-05
Ensure Token definition (XML) files are updated from the server

JamesSmartCell:
Patch to ensure token definition files are periodically checked for updates on the repo server.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/entity/FileData.java
+++ b/app/src/main/java/io/stormbird/wallet/entity/FileData.java
@@ -1,0 +1,15 @@
+package io.stormbird.wallet.entity;
+
+public class FileData
+{
+    public Long fileDate;
+    public String fileName;
+    public boolean modified;
+
+    public FileData()
+    {
+        fileDate = 0L;
+        fileName = "";
+        modified = false;
+    }
+}
--- a/app/src/main/java/io/stormbird/wallet/service/AssetDefinitionService.java
+++ b/app/src/main/java/io/stormbird/wallet/service/AssetDefinitionService.java
@@ -3,6 +3,7 @@ package io.stormbird.wallet.service;
 
 import android.content.Context;
 import android.os.Environment;
+import android.text.format.DateUtils;
 
 import org.xml.sax.SAXException;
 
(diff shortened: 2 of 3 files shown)


hboon:
There's this in the [README](https://github.com/alpha-wallet/contracts/blob/master/README.md):

> When a user accesses a contract through its Ethereum address, the user's agent (mobile client) downloads or updates the XML files from this repo.
> 
> “access” happens when:
> 
> * the app discovers that the user has (or had) tokens under a Contract;
> * the user actively seek to use a Contract (scan its QR code or copying its address from the Internet);
> * the user gets sent Universal Link which refers to a Contract or its assets
> * the app launches

colourful-land:
If you overlooked this part:
> Include the IF-Modified-Since header HTTP header with the local XML's last modified date in the GET request. If the XML on the server has not been modified, a 304 will be returned with an empty body.

Then you must have overlooked this part too:
> Be careful with timezone! Put a bit of thinking cases like this: the user moves from timezone A to timezone B and missing a latest update for up to 24 hours.

JamesSmartCell:
Code is tested against files on FTP and the lastmodified header is honoured in those that support it. It should start working when the server fills the header in correctly. Would be preferable to use a hash system (which I believe is one of the RFC standards) but file time will be an improvement over what is there now.

JamesSmartCell:
> Can you use a method not using the hourly timer? That look like the source of some evil.

See issue

JamesSmartCell:
Ready for merge - please revisit

JamesSmartCell:
Good idea.

JamesSmartCell:
OK implemented @hboon suggestion. If all good then I'll upload a release candidate for testing & release for China trip.

hboon:
Oh. Probably not in the scope of the PR, but I presume this is already handled already:

> When a user accesses a contract through its Ethereum address, the user's agent (mobile client) downloads or updates the XML files from this repo.
> 
> “access” happens when:
> 
> * the app discovers that the user has (or had) tokens under a Contract;
> * the user actively seek to use a Contract (scan its QR code or copying its address from the Internet);
> * the user gets sent Universal Link which refers to a Contract or its assets
> * the app launches

JamesSmartCell:
Yep those events trigger search for XML definition from Repo server.

PR #334, opened 2018-09-24
Handle empty repo gracefully

colourful-land:
2 changes.

- empty repo or starting repo server without giving repo path is handled in a graceful way.
- a last minute change to remove reference to https://awallet.io


dmz/src/main/java/io/stormbird/token/web/AppSiteController.java
```diff
@@ -67,7 +68,7 @@ public String getAppleDeepLinkConfigure() {
 
     @GetMapping("/")
     public RedirectView home(RedirectAttributes attributes){
-        return new RedirectView("https://awallet.io");
+        return new RedirectView("http://alphawallet.io");
```
hboon: More efficient to go to https://alphawallet.io instead, but not that important.

PR #344, opened 2018-09-26
Remove staging a-wallet.xml strings and merge into main strings.xml

JamesSmartCell:
- Remove old staging strings - awallet.xml (created to speed development and avoid confusion when we had a trust repo running in parallel with our own).
- Merge all strings into a single file and remove redundant strings.


James-Sangalli:
@JamesSmartCell conflict still exists. Please ensure this is fixed else we will have translation problems in the future.

PR #350, opened 2018-09-27
Consolidated token service

JamesSmartCell:
Merge disparate usages of Token maps into one for efficiency and code de-duplication.


James-Sangalli:
Think @hboon might have a good input on this one, I am having a bit of difficulty understanding what this PR is trying to achieve....

hboon:
@James-Sangalli it looks like there's not much code removed because the duplication is in the token object instances, not the code. Look for `tokenMap` in the diff. They are collections of tokens. There's a few of them in various classes affected by the diff.

With this PR, `RepositoriesModule` creates an instance of `TokensService` which is basically a collection of tokens, and pass it around, sharing just 1 copy of the token collection (hence part of the diffs being `tokenMap` -> `tokensService`). Other than memory savings, I reckon you can update (maybe pull them over the network or enrich the token data,) the shared, sole instance of `TokensService` and magically reap the rewards in other parts of the app with some reactive voodoo.

Sounds good :)

PR #352, opened 2018-09-27
Fix signature length correction

JamesSmartCell:
Discovered an old bug - sometimes there can be a lot of leading zeros in the signature. This should come as no surprise considering how bitcoin works. It doesn't happen often ...

Fixed.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/entity/SignaturePair.java
+++ b/app/src/main/java/io/stormbird/wallet/entity/SignaturePair.java
@@ -160,7 +160,7 @@ public class SignaturePair
             System.arraycopy(signature, 0, sigCopy, offset, 65-offset);
             for (int i = 0; i < offset; i++)
             {
-                sigCopy[0] = 0;
+                sigCopy[i] = 0;
             }
             signature = sigCopy;
         }
--- a/app/src/main/java/io/stormbird/wallet/entity/SignaturePair.java
+++ b/app/src/main/java/io/stormbird/wallet/entity/SignaturePair.java
@@ -153,11 +153,15 @@ public class SignaturePair
         //Now convert sig back to Byte
         signature = sigBi.toByteArray();
 
-        if (signature.length == 64)
+        if (signature.length < 65)
         {
+            int offset = 65 - signature.length;
             byte[] sigCopy = new byte[65];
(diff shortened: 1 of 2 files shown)


James-Sangalli:
@JamesSmartCell when is a signature greater than 65 bytes?

JamesSmartCell:
Good question and well spotted.

It's a phenomena of Two's Compliment. If the value of the final two bytes is greater than 0x80 then the BigInteger byte interpreter assumes a Two's Compliment value, and to keep the byte conversion positive it has to add an extra byte to the value to keep it matching. This breaks the 65 byte manipulation in the signing system. So, if the converted value is greater than 65 we know that BI has converted it into a Two's Compliment value and we just convert it back into the byte value if it were unsigned.

You may also wonder - say if the top two bytes were zero, and the next byte is greater than 0x80, it will grow into the final byte and break the signature system? It doesn't matter because the signing system converts the byte value back to BI and knows the value is unsigned, so it comes out the same.

app/src/main/java/io/stormbird/wallet/entity/SignaturePair.java
```diff
@@ -153,11 +153,15 @@ public SignaturePair(String qrMessage, String timeMessage, String contractAddr)
...
+        if (signature.length < 65)
         {
+            int offset = 65 - signature.length;
             byte[] sigCopy = new byte[65];
-            System.arraycopy(signature, 0, sigCopy, 1, 64);
-            sigCopy[0] = 0;
+            System.arraycopy(signature, 0, sigCopy, offset, 65-offset);
+            for (int i = 0; i < offset; i++)
+            {
+                sigCopy[0] = 0;
```
hboon: Maybe should be:

    sigCopy[i] = 0;

?

JamesSmartCell: Yes it should be. Thanks :) and thanks for looking, even if I'm not sure how you got here. Yes the buffer probably does initialise implicitly to zero, but for a signature it's best to be completely sure.

Be careful of making yourself too useful we don't want you to burn yourself out.

JamesSmartCell:
so ... do the lords deem this request now worthy of inclusion?

PR #356, opened 2018-09-29
Mark terminated contracts

JamesSmartCell:
This PR builds on the Token Service PR to identify terminated contracts and allow further optimisation of network usage. Once contracts are identified as terminated they are no longer updated or polled.


James-Sangalli:
@JamesSmartCell makes sense but need to fix conflicts

Issue #190, opened 2018-05-23
Add  a mechanism to collect the crash reports 


bootingman:
Hi, Sangelli,

two users report of crashing when tap AlphaWallet Android app logo after first-time installed on last night,I have collect some info  for reference.

1. Android 8.0 Huawei P10 Plus
2. collecting by asking

If we can't find how they crashed maybe we need to use 3rd party tools #282

James-Sangalli:
duplicate of #282

colourful-land:
I am trying to find the crash reports from Google:
https://play.google.com/apps/publish/?account=8548546370386799497#AppDashboardPlace:p=io.stormbird.wallet&appid=4973272515950350768

colourful-land:
> If we can't find how they crashed maybe we need to use 3rd party tools #282

Yes, we need to. Mark this as a duplicate of #282

PR #361, opened 2018-10-12
Firebase Crashlytics

justindg:
Added crash monitoring function via Firebase Crashlytics - was intend to solve #190

ACTION ITEM:
I would need a list of Google emails so I could add/transfer ownership of the dashboard. Currently only I have access to this.

Have tested with forced crashes via `Crashlytics.getInstance().crash()`. Screenshot below:

screen shot 2018-10-12 at 10 57 47 pm

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,22 +1,6 @@
-buildscript {
-    repositories {
-        maven { url 'https://maven.fabric.io/public' }
-    }
-
-    dependencies {
-        classpath 'io.fabric.tools:gradle:1.25.1'
-    }
-}
 apply plugin: 'com.android.application'
-//apply plugin: 'io.fabric'
 apply plugin: 'realm-android'
-
-repositories {
-    maven {
-        url 'https://maven.fabric.io/public'
-    }
-}
-
+apply plugin: 'io.fabric'
 
 android {
     compileSdkVersion 27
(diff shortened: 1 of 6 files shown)


hboon:
Does it work in China?

justindg:
@hboon Unfortunately I could not verify this. I've been seeing mixed opinions online. Fabric claims (via Twitter) that it works in China. Another Firebase employee (via Reddit) claims that it works as well.

colourful-land:
@justindg  Please add these people:

james.brown@alphawallet.com
weiwu.zhang@alphawallet.com
feedback@alphawallet.com

colourful-land:
@hboon 

a) if it crashes in China and we don't get a crash report -> that's unfortunate but acceptable (until we find a meeker provider)

b) if it starts the app by requesting a positive 'can-go' signal from Google (which is entirely blocked for not meek enough) and hangs the app there waiting for it, then it is not acceptable. Many websites are not accessible in China simply because it required one line of CSS from forbidden websites like wordpress.com and the webpage, unfortunately, waited for that.

In short, it doesn't have to "work" in China - it must not get in the way of working in China.

Please read the annotation in build.gradle:


    // WARNING! WARNING!  WARNING!  WARNING!  WARNING!  WARNING!  WARNING!  WARNING!  
    // WARNING!  Don't add depdency on Google Play Services without consulting
    // WARNING!  The Chinese marketing team.
    // WARNING!  Many Chinese Android phones kill any app that users google gms
    // WARNING!  on site, like immune systems cleansing infections.
    // WARNING! WARNING!  WARNING!  WARNING!  WARNING!  WARNING!  WARNING!  WARNING! 

You added `com.google.gms:google-services:4.1.0` which I believe is GOOGLE PLAY SERVICE would usually cause the app to send a package to Google upon starting, blacklisting the user in an instant.

colourful-land:
@justindg please find a way to add crash report without using google play service. Chinese team informs me that adding play service library is begging for trouble since it will check if the current Google Play service needs to upgrade -> choking the user.

colourful-land:
@justindg if you think this crashing report tool which depends on Google Play Service is far superior than the next guy, I can do a special build instruction for it to be built only for the PlayStore version (PlayStore is not accessible from China).

justindg:
Hi all, I've brought back Fabric Crashlytics (which is Firebase Crashlytics's "predecessor" anyway). Although I could still not verify if it 100% works in China, I found that it does not require Google Play Services, which is a good sign.

justindg:
> Hi all, I've brought back Fabric Crashlytics (which is Firebase Crashlytics's "predecessor" anyway). Although I could still not verify if it 100% works in China, I found that it does not require Google Play Services, which is a good sign.

Forgot to link the PR: https://github.com/AlphaWallet/alpha-wallet-android/pull/368

justindg:
See https://github.com/AlphaWallet/alpha-wallet-android/pull/368

PR #363, opened 2018-10-15
add warnings to prevent Google Play service being added in the future…

colourful-land:
… by another developer

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,12 +1,23 @@
+// WARNING WARNING WARNING
+// don't add any additional things here without first search "China" in this file
+
 buildscript {
     repositories {
         maven { url 'https://maven.fabric.io/public' }
+        // needed for pincode Lollipin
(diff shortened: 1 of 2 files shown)


hboon:
FYI. We have a [Danger](https://danger.systems/js/) file at https://github.com/AlphaWallet/alpha-wallet-ios/blob/master/dangerfile.js that contains code to check for such things for iOS. It's just Javascript with convenient access to the code changes. So we have simple checks like if a string is added to the en localisation, we at least warn that other localisations should be added for other messages.

It can be configured so it's run automatically with Travis for PRs. We haven't done that yet, so I'm just running it as a post-commit hook locally:

```
#!/bin/sh
yarn run danger local
```

colourful-land:
@hboon very helpful. Will investigate. Not high priority for now.

colourful-land:
Obsolete by PR #367

Issue #369, opened 2018-10-16
I buy a cryptokitties with alpha-wallet-android but failed 


duncanwang:
I buy a cryptokitties with alpha-wallet-android but failed. The transaction address is
https://etherscan.io/tx/0xe335dfe9717b84c89b9dc36ccf0c4c231f98dc0b67a82884bb7576bd33f178cd.
I find the cause is the Gas Price is too small, but there is no way to modify it.

PR #371, opened 2018-10-17
Implement gas settings dapp browser

JamesSmartCell:
As requested in issue #369, added confirmation activity that allows user to change the gas settings. 

- ensure we're using the current gas settings queried from the node.


James-Sangalli:
@JamesSmartCell is this getting the gas estimate for the transaction?

James-Sangalli:
@colourful-land bump

PR #374, opened 2018-10-24
Add ERC721 Support + UI enhancements

justindg:


JamesSmartCell:
- Required mods for ERC721 compatibility.
- Note that Travis failure is due to inability to retrieve a specific library. Reports of this failure are common, no accepted resolution.

PR #385, opened 2018-10-31
Fix opensea refresh

JamesSmartCell:
Fix for some tokens retrieved from Opensea not declaring they're ERC721, in addition ensuring the balance of all ERC721 tokens is refreshed correctly.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/service/TokensService.java
+++ b/app/src/main/java/io/stormbird/wallet/service/TokensService.java
@@ -6,8 +6,10 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 import io.reactivex.Observable;
-import io.reactivex.ObservableSource;
+import io.stormbird.wallet.entity.ERC721Token;
 import io.stormbird.wallet.entity.Token;
+
+import static io.stormbird.wallet.C.ETHER_DECIMALS;
 
 public class TokensService
 {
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/stormbird/wallet/service/TokensService.java
```diff
@@ -36,16 +38,24 @@ public Token addToken(Token t)
         return t;
     }
 
+    public Token addTokenUnchecked(Token t)
```
James-Sangalli: Naming is confusing

app/src/main/java/io/stormbird/wallet/service/TokensService.java
```diff
@@ -133,4 +142,28 @@ public void setCurrentNetwork(int currentNetwork)
     {
         this.currentNetwork = currentNetwork;
     }
+
+    public List<Token> getAllClass(Class<?> tokenClass)
```
James-Sangalli: Naming

JamesSmartCell:
It's related to an issue that was closed before. I will break it down for you.

During testing we noticed there were problems with the return from opensea. 

Some tokens explicitly declare they are ERC721 in the schema field. Some tokens do not.

When we first started looking at Opensea on Android all the tokens we tested with in the initial setup explicitly declared they were ERC721, which implies there are other schemas the API returns. So, we didn't accept anything that wasn't ERC721.

It turns out this is a param you can ignore - any token returned from opensea is ERC721. 

I guess Boon ran into the same issue at some point and fixed it by ignoring the schema.

This PR is an extension of that previous PR, since we need to account for changing balances of the tokens.

This PR is required for the next release.

hboon:
Right. On iOS, we ignore the `type` field (can't remember the exact name) in API results which often says `ERC721`, but not always.

app/src/main/java/io/stormbird/wallet/service/TokensService.java
JamesSmartCell: it gets all tokens of a specific class. Check the return and input types - it takes a class spec and returns a filtered list of Tokens. 

Maybe 'filterByClass' is a better name although inconsistent with the other methods.

PR #392, opened 2018-11-06
Improve refresh of wallet balances

JamesSmartCell:
Sometimes API return to check the wallet balances takes a while. 

- Updates the UI each time a value is received. 
- Check for null return from the API just in case of time-out etc.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/ui/widget/adapter/WalletsAdapter.java
+++ b/app/src/main/java/io/stormbird/wallet/ui/widget/adapter/WalletsAdapter.java
@@ -95,7 +95,10 @@ public class WalletsAdapter extends RecyclerView.Adapter<BinderViewHolder> {
     {
         for (Wallet wallet : wallets)
         {
-        	wallet.setWalletBalance(balances.get(wallet.address));
+        	if (balances.containsKey(wallet.address))
+			{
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/stormbird/wallet/ui/widget/adapter/WalletsAdapter.java
```diff
@@ -95,7 +95,10 @@ public void updateWalletBalances(Map<String, BigDecimal> balances)
     {
         for (Wallet wallet : wallets)
         {
-        	wallet.setWalletBalance(balances.get(wallet.address));
+        	if (balances.containsKey(wallet.address))
+			{
```
James-Sangalli: Formatting is slightly off here

app/src/main/java/io/stormbird/wallet/viewmodel/WalletsViewModel.java
```diff
@@ -204,6 +204,7 @@ private void updateBalances()
     private void updateList(Token token)
 	{
```
James-Sangalli: and here

JamesSmartCell:
ok. thanks for checking it. It's weird it looked ok in AS. I refreshed the formatting in both files just to be on the safe side.

PR #407, opened 2018-11-11
Use ens addresses

JamesSmartCell:
Use (Universal) ENS addresses in the send activity. These are looked up on the universal ENS directory on mainnet.


James-Sangalli:
@JamesSmartCell conflicts

PR #418, opened 2018-11-14
Fix import tokens interface issue

JamesSmartCell:
This PR ensures that we're using the correct Interface before doing transfers/trade and import tokens. ERC875 tokens could be using the older interface spec.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/entity/Token.java
+++ b/app/src/main/java/io/stormbird/wallet/entity/Token.java
@@ -466,6 +466,7 @@ public class Token implements Parcelable
 
     public void setInterfaceSpec(int b) { }
     public boolean isOldSpec() { return false; }
+    public boolean unspecifiedSpec() { return false; }
     public void setInterfaceSpecFromRealm(RealmToken ordinal)
     {
 
--- a/app/src/main/java/io/stormbird/wallet/entity/Token.java
+++ b/app/src/main/java/io/stormbird/wallet/entity/Token.java
@@ -262,6 +262,11 @@ public class Token implements Parcelable
         return null;
     }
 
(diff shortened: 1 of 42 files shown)


app/src/main/java/io/stormbird/wallet/service/TransactionsNetworkClient.java
```diff
@@ -197,4 +200,42 @@ private String readTransactions(NetworkInfo networkInfo, String address, String
...
+						{
+							TransactionContract ct = tx.operations[0].contract;
+							result = ct.decimals;
+							break;
+						}
+					}
+				}
+			}
+			catch (JSONException e)
+			{
```
James-Sangalli: Both silently fail!

app/src/main/java/io/stormbird/wallet/entity/Token.java
```diff
@@ -485,4 +507,19 @@ public boolean requiresAuxRefresh()
...
+    }
+
+    public boolean checkBalanceChange(Token token)
+    {
+        return !getFullBalance().equals(token.getFullBalance());
+    }
+
+    public int[] getTicketIndicies(String ticketIds)
+    {
+        return new int[0];
```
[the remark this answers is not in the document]
JamesSmartCell: Typo: yeah, well spotted! This is an intentional abstraction as nothing is supposed to happen when this function is called on the base class. You're right there should be a note that it's intentional.

app/src/main/java/io/stormbird/wallet/service/TransactionsNetworkClient.java
JamesSmartCell: correct. This happens, and we want a silent fail when we get a bad return from etherscan which can happen due to no fault of the app.

PR #423, opened 2018-11-15
Dynamic token refresh

JamesSmartCell:
This PR ensures the dynamic fields of the token cards are displayed correctly and refreshed on app startup and when the user swipe refreshes the tokens.


James-Sangalli:
@JamesSmartCell build is broken

PR #429, opened 2018-11-17
Bump version number for Release 1.1.0

JamesSmartCell:
Update release version and take a snapshot of code for Release 1.1.0

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,7 +38,7 @@ android {
         minSdkVersion 18
         targetSdkVersion 28
         versionCode 22
-        versionName "1.1.0"
+        versionName "1.30.0"
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
         multiDexEnabled = true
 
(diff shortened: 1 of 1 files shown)


JamesSmartCell:
OK, now I get why you called it 1.3

I need to do a new release to fix the issues Victor picked up anyway so I'll bump it in that.

JamesSmartCell:
This PR still needs to go through; it's a snapshot of version 1.1.0 which was released.

I'll prepare a release 1.3.0 which has the requested mods from Victor.

hboon:
Note that it’s 1.30.0 not 1.3.0 if we want to match the numbers.

hboon:
Note that it’s 1.31.0 not 1.3.0 if we want to match the numbers.

JamesSmartCell:
To stay in sync we need to have an intermediate release and then do version 1.31.0 once the Namespace code is in and working. Otherwise it doesn't make sense. We're matching the version numbers for the repo server to correctly serve the new files.

I will release 1.30.0 with the intermediate updates and then 1.31 once the namespace code is in.

hboon:
Yes that makes sense. (But the repo server still ignores version numbers, for now at least. You need to switch the repo server URL to the new domain if you want the new XMLs. Just mentioning this in case the code relies on this assumption.)

PR #437, opened 2018-11-19
Fix DApp contract construction

JamesSmartCell:
- Calculate gas required for transactions when in confirmation model.
- Fix Dialog box to give error message received from ethereum node.


James-Sangalli:
@JamesSmartCell is this more accurate than getting the gas price from the latest block?

Issue #322, opened 2018-09-09
Switch locale prompt's title is not localized


hboon:
![img_0215](https://user-images.githubusercontent.com/56189/45261304-ea2cc580-b431-11e8-9fa3-4259410a9cf6.JPG)

miang719:
@James-Sangalli pending for your translation strings

bootingman:
PR sent
https://github.com/AlphaWallet/alpha-wallet-android/pull/483

Issue #321, opened 2018-09-09
Backup wallet screen is not completely localized


hboon:
![img_0217](https://user-images.githubusercontent.com/56189/45261299-db461300-b431-11e8-92d6-d4f035b0e0c4.JPG)

miang719:
@James-Sangalli pending for your translation strings

James-Sangalli:
@miang719  up to @bootingman

bootingman:
PR sent 
https://github.com/AlphaWallet/alpha-wallet-android/pull/483

Issue #320, opened 2018-09-09
Add/Import wallet prompt is not localised


hboon:
![img_0216](https://user-images.githubusercontent.com/56189/45261291-a9cd4780-b431-11e8-9c18-726c8c9adbfe.JPG)

miang719:
@James-Sangalli pending for your translation strings

James-Sangalli:
@JamesSmartCell Chinese is not included in the localisation service as Chinese and English were the first two languages of the app

James-Sangalli:
These will be included in the translation service which will happen in clumps. closing all these issues.

PR #439, opened 2018-11-21
update the Chinese strings

bootingman:
1. add strings from /values/strings.xml into /values-zh/strings.xml
2. translated the strings
3. replace a string "Trust Wallet" with "AlphaWallet"

### DIFF
--- a/app/src/main/res/values-zh/strings.xml
+++ b/app/src/main/res/values-zh/strings.xml
@@ -1,16 +1,98 @@
 <resources>
+
     <string name="app_name">AlphaWallet</string>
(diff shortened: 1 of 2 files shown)


JamesSmartCell:
Is this derived from the 'update-translations' branch? If not please use that one as the working branch. I balanced the strings between all the languages so that any inconsistency will show up immediately.

James-Sangalli:
For #320 #321 & #322

James-Sangalli:
I am gonna add Chinese to the translation service, it is simply more efficient to have them do it. Closing.

Issue #357, opened 2018-10-02
Enable SNS notifications 


James-Sangalli:
This has already been setup on iOS and has the control panel setup for SNS in general. The android version is not implemented in this repository yet.

Requires the following:
- Be able to register the device token to SNS
- Be able to get a notification from the control panel
- Enable the appropriate permissions
- Work with Baidu cloud (can be pursued after GCM or Firebase) 

I have credentials for SNS and firebase which I can provide on demand.

James-Sangalli:
@justindg this boilerplate code might help you: https://github.com/AlphaWallet/alpha-wallet-android/tree/with-amazon-sns

miang719:
@justindg, any follow up on this issue?

justindg:
Hi all, sincere apologies for the delay - I am able to send Push Notifications via Firebase already but I'm running into issues integrating with SNS.

Some concern though: a few weeks back we've talked about the issue with Google Play Services and China. Firebase Messaging requires Google Play Services for it to work. @James-Sangalli, what did you use for iOS?

James-Sangalli:
@justindg the firewall problems are not an issue for iOS because iOS has it's own notification centre. 

could you be more specific about your issues with SNS?

justindg:
@James-Sangalli I am now able to register device tokens and publish notifications to SNS. 

Just wondering - how/where did you store AWS credentials? Figured they're needed to register the device tokens automatically through the app.

Issue #282, opened 2018-07-17
Use third party crash analytics 


James-Sangalli:
https://www.captechconsulting.com/blogs/a-comparison-of-android-crash-reporting-tools mentioned Firebase, Crashlytics, Apteligent, Bugsnag

colourful-land:
@justindg  do you have practical experience with any 3rd party crash analytics tool?

justindg:
I've had a nice time with Firebase Crashlytics! Easy to setup too.

miang719:
@justindg just do it then. thanks

miang719:
@justindg how long does it take to setup? so i can follow up the progress

justindg:
@miang719 done. PR can be found here https://github.com/AlphaWallet/alpha-wallet-android/pull/361

I would need a list of Google accounts so I could add new members/owners for dashboard access. Thanks!

colourful-land:
Thanks a lot. I'll hold off for a day or two until I figure out our main domain name for google accounts.

colourful-land:
@bootingman @zhangzhongnan928 

Tell a friend in China to download this:

https://github.com/AlphaWallet/alpha-wallet-android/releases/tag/1.0.11

Manipulate the basic functions (e.g. restore from a backup key and check its balance, send an ether to or back).

If it crashes, let us know. If it does not crash, force a crash with the following:

Settings > Help & FAQ > Click anywhere on the blank space below the FAQs (it will mysteriously crash)

hboon:
@colourful-land does it need to be launched one more time so the app can report the crash to the server?

bootingman:
@colourful-land @hboon I'm sure, several crash happened in Zhejiang province, China

colourful-land:
I didn't see any crash reports in firebase, so I believe the crash report might have been sent to  Fabric Crashlytics. Please have a look @justindg if there are new crash reports of today and note down their IP addresses. If it has worked as expected we still can release it today.

† https://console.firebase.google.com/project/alpha-wallet-android-186eb/overview

colourful-land:
Fits the profile. Problem solved. (I'll ask for login credential separately by email.)

PR #440, opened 2018-11-22
Allow push notifications through SNS and FCM

justindg:
It is important to note that push notifications through SNS and FCM requires Google Play Services. As we have discussed before in #282,  there is a probability that the app would not work in China because of this.
 
Nevertheless, this PR satisfies the requirements in #357 
- Be able to register the device token to SNS
- Be able to get a notification from the control panel
- Enable the appropriate permissions


JamesSmartCell:
TODO: Check that this works in China. @bootingman

PR #451, opened 2018-11-29
Update Web3 provider

JamesSmartCell:
Update to the latest web3 provider found here:

AlphaWallet/AlphaWallet-web3-provider/dist

Also remove redundant code and fix the 'signPersonalMessage' prompt to display the data to be signed in text rather than hex string.

TODO: will investigate how to update from github on file change using gradle.


app/src/main/res/raw/init.js
```diff
@@ -61,3 +56,4 @@ window.web3.eth.getCoinbase = function(cb) {
     return cb(null, addressHex)
 }
 window.web3.eth.defaultAccount = addressHex
+window.web3.currentProvider.isAlphaWallet = false
```
hboon: We should probably set `isAlphaWallet` to `true` instead. Right @James-Sangalli  ?(iOS sets it to false too).

```
web3.currentProvider.isAlphaWallet = true
```

JamesSmartCell: Yes, I thought the same thing.

Also, I notice we set isNeuron=true in the main provider code.

Maybe move the isAlphaWallet out into the main code and delete from the init code.

JamesSmartCell: I wonder if anyone is checking for alphawallet? If not I'll add a check on the office door, or at least I would if I was able to flash the firmware. I didn't add an OTA flash facility.

hboon: So @JamesSmartCell , you can use this new commit for the updated provider which sets `isAlphaWallet` to `true`: https://github.com/AlphaWallet/AlphaWallet-web3-provider/commit/f25206c50009d1eb922c3cc8c0ba91594155e8b6

JamesSmartCell:
Provider updated from the web3-provider repo.

PR #456, opened 2018-12-03
Fix transaction view crash

JamesSmartCell:
- Fix for item observed on Crashalytics.
- Fix the date headers in the transaction view so transactions appear on the correct date in the phone's local time zone.


James-Sangalli:
@JamesSmartCell please reference the crashanalytics error

PR #478, opened 2018-12-24
Iframe display

JamesSmartCell:
* Note: this PR also contains the 'use-cached-balance' PR, so handle that one first to see the correct content in this PR. *
 
Update IFrame display to reflect latest spec as per 20/12:

- AssetDetail shows the 'introduction' content for zero balance.
- AssetDetail shows the 'instruction' content for positive balance.

### DIFF
--- a/lib/src/main/java/io/stormbird/token/tools/TokenDefinition.java
+++ b/lib/src/main/java/io/stormbird/token/tools/TokenDefinition.java
@@ -1,5 +1,6 @@
 package io.stormbird.token.tools;
 
+import javax.swing.text.html.HTML;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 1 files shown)


JamesSmartCell:
OK leaving as a WIP means that I will need to pull this branch into the current code - 

This means that future branches will include the source in this one so the current state of this branch will be included in any future PR. Not sure that's within git's normal usage; Will it look confusing in the review of the next branch that includes the current state of this?

JamesSmartCell:
Releases depend on this code. So, each subsequent branch that includes this code will be 'infected' with the WIP tag. Configuration control is going to become very complicated. 

I already have a branch that is dependent on this one, so that now has to WIP as well. When eventually all these branches are merged into master it's going to be the mother-of-all pull requests!

hboon:
Generally, the approach would be to create a separate PR that contains the commits that needs merging into master and merge that PR, and leave the bits that are still WIP in this PR [1]. Any fixes to the iframe bits can continue in this branch. Looks like this PR is to display iframe, are there code that depends on iframe in production?

[1] And after that 2nd PR is merged into master, rebase this PR's branch against master, which will rewrite the commits of the branch so that it is a fork of master from the HEAD again. As there are new PRs merged into master, just continue (or not) rebase this PR's branch against master.

JamesSmartCell:
OK I understand what you're saying. I think I would have to have started the branch with that intention then I would have crafted the branch tree differently. I didn't do that.

Hence - I would rather this branch was merged into master and I can start a 'IFrame-Experimental' WIP branch that feeds into the main one once what we want is ready.

PR #500, opened 2019-01-11
Fix transaction placeholder

JamesSmartCell:
Ensure transaction placeholder 'shield' doesn't appear when in offline mode.


James-Sangalli:
@JamesSmartCell don't understand the purpose of this PR

JamesSmartCell:
There's a nice graphic pop-up that's meant to appear when there's no transactions on an account. It was appearing (incorrectly) when there were transaction on an account, but the user's phone was not connected to internet. It's cosmetic.

PR #518, opened 2019-01-26
Fix for differing timezone usages in XML file

JamesSmartCell:
This PR fixes an issue with Timezone usage in TBML files. If time is given without timezone information, the library should not just throw an exception. This is useful for development but not useful elsewhere.

The code was changed to make a demo possible which used the FIFA world cup tickets. There is another possible fix where any time used in any token is given in the XML file.

If there's no explicit time given in the XML file, any time read should be considered to be UTC, and displayed in the user's current time zone, which is what this patch amends.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/entity/Ticket.java
+++ b/app/src/main/java/io/stormbird/wallet/entity/Ticket.java
@@ -9,6 +9,8 @@ import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
+import io.stormbird.token.util.DateTime;
+import io.stormbird.token.util.DateTimeFactory;
 import org.web3j.abi.TypeReference;
 import org.web3j.abi.datatypes.Function;
 import org.web3j.abi.datatypes.Type;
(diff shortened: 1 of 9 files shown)


James-Sangalli:
From @colourful-land "If the data is UnixTime(BinaryTime in XML), it's 4-bytes and based on UTC, and the code to handle that shouldn't appear in ZonedTime class.

If the data is GeneralisedTime (at least 12 bytes in ASCII format), it's based on user's local timezone. In that case, ZonedTime class should be patched up to default to user's timezone anyway, and not throw an exception.

So you need to touch 2 places. Thanks!"

JamesSmartCell:
thanks. I already discussed it with Weiwu, changes will be forthcoming (but slightly different from what is outlined above - as that doesn't appreciate what the actual issue is).

colourful-land:
Dear @JamesSmartCell in your branch it has:
> Latest commit 3094a63  25 days ago

Which means this isn't handled, correct?

> "If the data is UnixTime(BinaryTime in XML), it's 4-bytes and based on UTC, and the code to handle that shouldn't appear in ZonedTime class.

> If the data is GeneralisedTime (at least 12 bytes in ASCII format), it's based on user's local timezone. In that case, ZonedTime class should be patched up to default to user's timezone anyway, and not throw an exception.

> So you need to touch 2 places. Thanks!

JamesSmartCell:
Coincidentally I was working on this while you were looking at it.

I've finished it in a more Java way and I think it makes the code a bit cleaner.

It's urgent to merge or suggest other changes because the website (DMZ) depends on these fixes to render tickets that are live.

James-Sangalli:
@JamesSmartCell a rebase would be good too :)

James-Sangalli:
@colourful-land let's hurry to get this reviewed, it is messing up the landing page.

JamesSmartCell:
yes tested with app and website.

PR #519, opened 2019-01-26
Fix URLs for xDAI network

### LINKED ISSUE #515: xDai don't pick up the transection records
image

JamesSmartCell:
Correct the definitions for API URLs so the history and transaction/contract lookup works correctly.

Fixes #515 and #516


James-Sangalli:
@JamesSmartCell probably should make these into Constants... Seems to be referenced multiple times across many files...

PR #520, opened 2019-01-26
Fix for bad gas price value

### LINKED ISSUE #514: Need to set the default gas fee for xDai Chian at 1 Gwei
image

image

JamesSmartCell:
- Ensure we don't use false gas price updates. DAI node often returns zero gas price which is an error.
- Update gas price with live value on confirmation activity page.

fixes #514

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/interact/FetchGasSettingsInteract.java
+++ b/app/src/main/java/io/stormbird/wallet/interact/FetchGasSettingsInteract.java
@@ -1,6 +1,7 @@
 package io.stormbird.wallet.interact;
 
 
+import android.arch.lifecycle.MutableLiveData;
 import io.stormbird.wallet.C;
 import io.stormbird.wallet.entity.GasSettings;
 import io.stormbird.wallet.repository.GasSettingsRepositoryType;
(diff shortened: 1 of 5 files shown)


James-Sangalli:
@JamesSmartCell why the need for an interval? Don't you simply get the gas price on demand?

James-Sangalli:
@JamesSmartCell initiate send screen -> load gas price from node -> user may override after its been loaded on the initial screen, please clarify my misunderstanding as I cannot see how that conflicts with the user's choice.

James-Sangalli:
@JamesSmartCell xdai network has a fixed fee of 1 GWEI

JamesSmartCell:
Your flow example assumes the user will push the transaction through immediately. In practice this may not be the case.

Flow: user wants to push transaction, tx settings page is opened with current gas price. Time passes, gas fee increases. With no live update, tx will go through with the settings used at the time. Alternatively perhaps gas fee decreases.

One alternative is for a gas fee refresh button. However this is an extra UI element which may clutter the interface and make things more complex appearing.

I requested Farrah to check out the metamask implementation which I believe is a good UI - with three buttons for slow/cheap, default and fast/expensive. These values potentially are updated live - would need to check the source code or leave the screen running and see if they change.

James-Sangalli:
@JamesSmartCell Ok, I think i get it now, just to clarify: this only happens when user is on the page, it doesn't enquire while idle?

PR #535, opened 2019-01-30
Fix gas default

JamesSmartCell:
This patch ensures the correct gas price default is set - currently we default to 1 Gwei for xDai and 30 Gwei for all others - if we don't hear otherwise. xDai will always be 1 Gwei unless (for whatever reason) the user wants to up the gas.


James-Sangalli:
@JamesSmartCell 30 gwei is quite high?

PR #538, opened 2019-02-01
removed unused code and files

James-Sangalli:
This PR tidies up all the unused code and files.

All was done via safe delete and runs cleanly but please have a spot check


James-Sangalli:
@JamesSmartCell builds are all running fine and I was able to play around with it on the emulator looking for things like images and labels to see if they are intact (all good from what I have seen)

the labels that were removed are only the ones that were never used in the first place, if a label id is used it would have stayed intact, therefore referencing wouldn't have changed as the ones that were implemented can still be referenced. 

Perhaps you would like to try it out on your phone first? From my angle it looks good, perhaps @colourful-land could also give it a spin.

James-Sangalli:
@JamesSmartCell let's merge this ASAP, please have a try on your phone then merge, it should work fine.

JamesSmartCell:
We'll push it through after the release tonight.

Issue #505, opened 2019-01-19
failed to import magiclink for spawnable token


lyhistory:
network: ropsten
meetup contract address: 0xa2918727EBD4B6c16cA3Bd25Becb0C739b70339d
organizer address: 0xEda3c41535ae147BD10d7B18884C7340b893cbDc
magic link: https://app.awallet.io/AgAAAABcx9Y4opGHJ-vUtsFso70lvssMc5twM51DTjIwMTkwNDMwMTI1OTM2KzA4MDAAAQAAAAAAAAAAQWG96vQO--XkIKyrC4QOSFZP__pZEEm-YMyxWov3L3LiOGpM23xgFtW1bq8Cc6213GdfBYtbbqi6YxKCkMzXmfsb

I can redeem it with my mockup tool using web3j:
![image](https://user-images.githubusercontent.com/1522513/51422959-47603100-1bf3-11e9-99f5-6af182356553.png)
https://ropsten.etherscan.io/tx/0x7f574894afd632c6943b5dcb9d87724fcc95c972a273fd483d0636b1a3d9b36f

while I open the magic link with the latest android client, it shows:
![image](https://user-images.githubusercontent.com/1522513/51423025-00bf0680-1bf4-11e9-955f-62c6d24f1e01.png)
but the client can display it nicely:
![image](https://user-images.githubusercontent.com/1522513/51423073-b427fb00-1bf4-11e9-9f81-b0801bc3916c.png)

please note that currently there's no validation on whether the magic link has been redeemed(in the meetup contract: spawnPassTo in https://github.com/AlphaWallet/contracts/blob/master/spawnable-contract/schema1/MeetupContract.sol), so we can redeem it multiple times (maybe we should add validation for this later on)

and this is how I extract tickets from the magic link for your reference:
![image](https://user-images.githubusercontent.com/1522513/51423110-72e41b00-1bf5-11e9-9bbc-fb25ccd5d767.png)

James-Sangalli:
@lyhistory is the signer address 0x52b590123a81a22f08acf05598fec68c636b4797?

James-Sangalli:
@lyhistory isn't your code using the same code as in android? @JamesSmartCell please take a look also

lyhistory:
hi, @James-Sangalli previously we have a discussion here: https://github.com/lyhistory/Peter-s-Tool/issues/15

[...]

PR #544, opened 2019-02-02
Add compatibility with spawnable and currency magic links

James-Sangalli:
For #505 and adds compatibility for future use of currency drop magic links

@JamesSmartCell I am leaving this as a WIP because I noticed that the master branch (which this is based off) is not working with regular magic links coming from iOS. 

Once that issue is fixed I will rebase this branch with master and retest with this branch, if it works I will remove the WIP label and we can merge it in.

### DIFF
--- a/lib/src/main/java/io/stormbird/token/tools/ParseMagicLink.java
+++ b/lib/src/main/java/io/stormbird/token/tools/ParseMagicLink.java
@@ -87,6 +87,7 @@ public class ParseMagicLink
      * bytes32: 30818B896B7D240F56C59EBDF209062EE54DA7A3590905739674DCFDCECF3E9B
      *
      */
+
     public MagicLinkData parseUniversalLink(String link) throws SalesOrderMalformed
     {
         final String importTemplate = "https://app.awallet.io/";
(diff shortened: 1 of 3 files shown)


lib/src/main/java/io/stormbird/token/tools/ParseMagicLink.java
```diff
@@ -92,81 +99,167 @@ public MagicLinkData parseUniversalLink(String link) throws SalesOrderMalformed
...
-            //now read signature
-            ds.readSignature(data.signature);
-            ds.close();
-        } catch (IOException e) {
-            throw new SalesOrderMalformed();
-        } catch (StringIndexOutOfBoundsException f) {
-            throw new SalesOrderMalformed();
         } catch (Exception e) {
+            //TODO handle properly
             throw new SalesOrderMalformed();
```
JamesSmartCell: SalesOrderMalformed is caught in the classes that use this method. It's purposefully abstracted this way.

James-Sangalli: Each method I have seen using this function only prints the stack trace... I should however move my comment to there instead so you're right.

James-Sangalli:
@JamesSmartCell I will rewrite the class based off of the iOS version here: https://github.com/AlphaWallet/alpha-wallet-ios/blob/master/AlphaWallet/Market/UniversalLinkHandler.swift then resync with @lyhistory so that he uses the same lib classes for his generators

Issue #542, opened 2019-02-01
Auto pickup below tokens on xDai network


hboon:
Refer to https://github.com/AlphaWallet/alpha-wallet-ios/issues/1075

hboon:
Save you some copy and pasting:

```
        (name: "DEN", contract: "0x6a814843de5967cf94d7720ce15cba8b0da81967"),
        (name: "BURN", contract: "0x94819805310cf736198df0de856b0ff5584f0903"),
        (name: "BURN", contract: "0xdec31651bec1fbbff392aa7de956d6ee4559498b"),
        (name: "BURN", contract: "0xa95d505e6933cb790ed3431805871efe4e6bbafd"),
        (name: "DEN", contract: "0xbdc3df563a3959a373916b724c683d69ba4097f7"),
        (name: "DEN", contract: "0x6e251ee9cadf0145babfd3b64664a9d7f941fcc3")
```

zhangzhongnan928:
Add one more, the same group ETHDever

https://blockscout.com/poa/dai/tokens/0x3e50bf6703fc132a94e4baff068db2055655f11b/token_transfers

Issue #468, opened 2018-12-14
Automatically pickup partner contracts


hboon:
Refer to https://github.com/AlphaWallet/alpha-wallet-ios/issues/961

PR #549, opened 2019-02-03
Check known contracts

JamesSmartCell:
Add a contract watch list system, currently for xDai contracts.

See issue #542


JamesSmartCell:
Ok #468 contracts added. I had to for now assume that these are ERC20 contracts, because the constructor scanner doesn't currently handle delegate contracts.

PR #552, opened 2019-02-03
Spawnable and xdai links compat

JamesSmartCell:
This is the Java native version of parsing byte arrays. The buffer is used for convenience. You could write your own serial byte handler class; it might be better than the java implementation (it should be more efficient at least). It's probably better to extend the existing one though.


James-Sangalli:
@JamesSmartCell  Test fails on this assert:                
 assertEquals(order.priceWei, data.price);

JamesSmartCell:
Added WIP label as should not be merged until API endpoints are in.

James-Sangalli:
@JamesSmartCell it's cool you added in the function sig for dropCurrency however in real life we must use the payment master. Could still be good for testing however.

James-Sangalli:
@JamesSmartCell can test the paymaster locally with this branch: https://github.com/AlphaWallet/PaymentServer/tree/add-xdai-magic-link-generator Haven't merged yet and it could change a bit with my testing, just a heads up.

JamesSmartCell:
Generating the links from within the app is only for testing, but the code in the standalone library is intended to be fitted into Peter's Tool and any servers written in Java that are using magiclink. Currently the code in Peter's Tool doesn't seem to work correctly; it generates links that don't decode correctly at the moment.

James-Sangalli:
@JamesSmartCell tried this magic currency link generated from android build: https://app.awallet.io/BFhEQUlEUk9QAAAACgAAJxBcXj9gbruzvdcbOXSf_WxABoSlLGHRejf1V4soe0YlTEVQdcWFM0sX2rByxqq61CDXfMJ9xy3FxxJIN9COdgnhNVH4l2I9wNFDKuduOFTWWdlAQsVvV1j-HA==

Android app doesn't recognise it and iOS has an issue with its signature (despite the byte length being correct), please point me in the correct direction.

JamesSmartCell:
Link generation is only done inside the library. Links are functioning with the contract that's in the repo. The spec matches the actual input to the dropCurrency function with a leading designation byte.

James-Sangalli:
@JamesSmartCell Why does this need to be network specific? A user may want to drop ropsten or even regular eth and we should allow that possibility.

James-Sangalli:
> Link generation is only done inside the library. Links are functioning with the contract that's in the repo. The spec matches the actual input to the dropCurrency function with a leading designation byte.

We need to be able to generate them easily, say through a main method call in a class, would you be able to do that for simplicity to the end user? (Potentially dai team for example)

JamesSmartCell:
You can't do the whole generation in one go as you need to get a signature. But, it could be done in three steps though instead of 4. Check the link formation in the test, it shows how to do it.

PR #576, opened 2019-02-12
Fix ERC20 Send

JamesSmartCell:
Fix to ensure ERC20 tokens can be sent correctly. Also display correct token data in ERC20 view header.


JamesSmartCell:
When the new UI was done, there's a certain flow path where if you're trying to send ERC20 tokens you would end up sending ETH instead. Not good!

PR #597, opened 2019-02-19
support multiple domains in magic links for android app

James-Sangalli:


app/src/main/java/io/stormbird/wallet/ui/ImportTokenActivity.java
```diff
@@ -261,20 +258,20 @@ private void onImportRange(TicketRange importTokens)
```
JamesSmartCell: Either use a local for the paymasterUrlPrefix, or use member consistently.

PR #611, opened 2019-03-02
Stability update

JamesSmartCell:
Stability update from testing


James-Sangalli:
@JamesSmartCell not clear what this PR is about

JamesSmartCell:
Look at the source code

JamesSmartCell:
It's fixes for things that are needed to be fixed.

colourful-land:
@JamesSmartCell looking at the code, it seems rather likely the change is intended to solve a bug, but I can't see what bug it is. Is it possible for you to give the problem before giving the solution?

PR #625, opened 2019-03-06
Wallet View token update fix

JamesSmartCell:
Simplify the wallet page token update.
- Ensure filters work 'live' (previously they didn't).
- Ensure tokens are updated as they are discovered (this was previously delayed until a refresh).

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/ui/WalletFragment.java
+++ b/app/src/main/java/io/stormbird/wallet/ui/WalletFragment.java
@@ -159,7 +159,7 @@ public class WalletFragment extends Fragment implements View.OnClickListener, To
 
     private void onToken(Token token)
     {
-        adapter.updateTokenCheck(token);
+        adapter.updateToken(token);
     }
 
     private void initTabLayout(View view) {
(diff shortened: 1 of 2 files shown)


James-Sangalli:
@JamesSmartCell confused as to what this PR is, please be more clear.

JamesSmartCell:
Sorry I thought the comment above was OK.

Ensure filters work 'live' (previously they didn't).

On the wallet page there are a set of filters, namely 'all', 'currency' and 'assets'. When the user clicks on one of these filters, the expected behaviour would be that the tokens are filtered accordingly. For example if the user clicks on 'currency' filter tab they would expect to see only token cards showing native currencies.

During testing, we noticed they didn't update live, there was a noticeable delay and usually you'd manually refresh to see the change.

This PR mainly fixes that issue.

The other issue that was noticed was that it was taking a while for the tokens discovered during the transaction sync to appear in the wallet view. This was down to the way the Recycle view works and required a little re-factor to get it working reliably.

Most of these issues were discovered during the Multi-Token support going in, but they benefit the wallet pre-multi-Token support and I felt the upgrade could be in its own PR, since it is focussed in one area.

PR #623, opened 2019-03-04
Rework Wallets Page to accommodate wallet names

justindg:
Here are the details of the rework:
- Wallet can now have names
- **Export** and **Delete** has been removed from the Wallets screen, replaced by **More** button
- **More** button takes the user to a new **Wallet Actions** screen, where user can rename, delete, backup wallet

 @farrahfazirah - Would love to hear feedback for the UX. Feel free to give recommendations to improve it

PR #627, opened 2019-03-07
Add wallet names to addresses

James-Sangalli:
Rebased version of #623 

if this version is okay, we can use it instead of the other.


JamesSmartCell:
Extra work was required on the main source tree as the activities needed a refactor to make them less complex

PR #650, opened 2019-04-04
Fix fragment removal

JamesSmartCell:
Fixed issue from crashlytics and add 'home' to menu to allow users to return to home page


JamesSmartCell:
requires slight flow refactor

PR #657, opened 2019-04-07
Fix payment server default check

JamesSmartCell:
use default payment server address check.


James-Sangalli:
@JamesSmartCell no idea what this PR means

JamesSmartCell:
It's a fix to use the correct payment server URL. Previously it was wrong under certain circumstance with spawnable on currency link.

James-Sangalli:
@JamesSmartCell for posterity please be more specific, what is the exact issue?

JamesSmartCell:
Spawnable link wouldn't have the /api/ extension, due to old code overriding it.

The wrong API was called to check if the paymaster will fund a certain contract, so the result would always be negative - even if the contract is funded the user would have to fund the transfer.

PR #664, opened 2019-04-15
Fix ENS resolution

JamesSmartCell:
Fix for ENS resolution. Previously used method was incorrect. Switch to using standard Web3j methods for validating ENS names and resolving address.


James-Sangalli:
@JamesSmartCell please reference the issue, #664

PR #666, opened 2019-04-15
Fix fragment issue

JamesSmartCell:
Use the correct 'onAttach' system to setup callbacks which guarantees we never have the situation where the fragment appears without setting up the callbacks.

Fix a few edge conditions resulting in crashlytics upsets.


justindg:
Nice catch on onAttach issue, totally missed that one. :)

PR #669, opened 2019-04-16
Make all relevant transactions appear in the token card view

JamesSmartCell:
Ensure relevant transactions always appear in the detail view.


James-Sangalli:
@JamesSmartCell unclear on the meaning of this PR, can you write clearly please?

JamesSmartCell:
Ensure transactions relevant to the user's wallet appear in the transaction detail view (the new view that was designed showing the short transaction history with send/receive at the bottom). I noticed during EDcon that when there's only occaisional activity on the account the transactions didn't always appear there, even though the balance updated. This is most likely because of a lag between infura sensing the balance update and etherscan logging the transaction. Best to play it safe and continually update the transaction list from the network when in focussed view.

PR #670, opened 2019-04-16
Handle Magiclink if imported via dapp browser

### LINKED ISSUE #659: If user open MagicLink from AlphaWallet DApp Browser, it should auto launch import screen

JamesSmartCell:
resolves #659

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/ui/DappBrowserFragment.java
+++ b/app/src/main/java/io/stormbird/wallet/ui/DappBrowserFragment.java
@@ -31,6 +31,8 @@ import android.widget.Toast;
 
 import com.google.gson.Gson;
 
+import io.stormbird.token.tools.ParseMagicLink;
+import io.stormbird.wallet.entity.*;
 import org.web3j.crypto.Keys;
 import org.web3j.crypto.Sign;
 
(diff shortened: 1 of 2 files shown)


hboon:
@JamesSmartCell This looks like it handles QR code scanning (nice). Does it handle these:

1. User types or pastes the MagicLink into the dapp browser URL bar and hits enter
2. User taps on a link in the current page in the dapp browser. The tapped URL is a MagicLink

(the issue title could have been clearer, sorry!)

JamesSmartCell:
User would have to try very hard to get the URL into the browser (I inserted extra characters into the URL to get round the clipboard sensor), but no it takes user to the website.

Is the desired behaviour to handle them as import? I guess it is; what occasion would a user ever want to see the web rendering of the token, and not be taken to the import page.

JamesSmartCell:
handling in the URL is good as it catches clicked links too. It's interesting that the web intercepts don't work in the internal browser.

hboon:
> handling in the URL is good as it catches clicked links too.

Yes.

> It's interesting that the web intercepts don't work in the internal browser.

Yeah, same behaviour on iOS, I think it's good though. Let's you have a chance to intercept and put in custom behaviours.

JamesSmartCell:
Ok behaviour updated, works on URL

hboon:
@JamesSmartCell looks good. Thanks. Let you do the merging since there's another PR. In case you want to verify or rebase master.

PR #676, opened 2019-04-21
Fix issues reported from Crashlytics

JamesSmartCell:
- Token update sanity check.
- Browser history item sanity check & back/forward logic update.


James-Sangalli:
@JamesSmartCell please write these messages clearly, it is not clear what this PR actually does.

JamesSmartCell:
It's a small PR that patches an issue reported on crashlytics (GPF), and fixes an inconsistency in the same code I noticed while exhaustively testing the fix. The back forward buttons on the dapp broswer didn't quite work consistently. The code was fixed so they work correctly and have tests at each stage in case of edge conditions. Real world use on multiple devices and OS's throws unexpected things at you. Ideally we should be testing on at least the major OS variants: Samsung Huawei Motorola and Xiaomi.

PR #683, opened 2019-04-27
Add tokenscript validator API to DMZ

JamesSmartCell:
Added tokenscript validator to allow iOS and Android to have identical signature check in DMZ. This is intended to provide a server-side workaround for #721

TODO: Validate signature is from recognised issuer.


JamesSmartCell:
Good idea. I'll expand the fail information.

JamesSmartCell:
Any suggestions for a standard way to verify the issuer? There must be an industry standard for how browsers verify signature issuer validity.

colourful-land:
Since the server runs Linux, I guess the issue is relatively simple - Linux just put the issuer list in /etc/ssl while Windows requires you to access it in a programmable way.

@jot2re can you hint how to do this? Should we just go through each of the pem files in /etc/ssl and see if the chain certificate is signed by one of them, or is there a default Java solution to check if a chain cert is signed by a known authority?

jot2re:
I am not sure that there is a default way of doing this, so I think you are right, that you need to go through each of the certificates in the root store. However, I don't think it is necessary to check the signature of each certificate. Instead I would just check the common name of each pem file and see if it matches the name of the key (or the fingerprint if you have access to that) and check the signature once you find a certificate with the right name. 
Also remember to verify that all the certificates in the chain are still valid.

colourful-land:
@James-Sangalli would be great to resolve the conflict & rebase first.

James-Sangalli:
For OSX

    keytool -list -keystore /Library/Java/Home/jre/lib/security/cacerts -v

colourful-land:
You need this to understand how the chain of certificate works. https://docs.oracle.com/cd/E19316-01/820-2765/gdzen/index.html

James-Sangalli:
@colourful-land @JamesSmartCell currently can check issuer against cacerts but noticed that it breaks when there is a DSA algorithm on signature.validate.

James-Sangalli:
> @colourful-land @JamesSmartCell currently can check issuer against cacerts but noticed that it breaks when there is a DSA algorithm on signature.validate.

@jot2re I believe the algorithm is defaulting to RSA somewhere in here: 
        DOMValidateContext valContext = new DOMValidateContext(new CertKeySelector(), nl.item(0));

in CertKeySelector, could you point me in the right direction to set the algorithm properly using the property inside the XML file?

jot2re:
> > @colourful-land @JamesSmartCell currently can check issuer against cacerts but noticed that it breaks when there is a DSA algorithm on signature.validate.
> 
> @jot2re I believe the algorithm is defaulting to RSA somewhere in here:
> DOMValidateContext valContext = new DOMValidateContext(new CertKeySelector(), nl.item(0));
> 
> in CertKeySelector, could you point me in the right direction to set the algorithm properly using the property inside the XML file?

Which XML file are you using? I have issues accessing Tokenscript.org. In any case, _very briefly_ looking at the code, it seems it will return the first public key it encounters, but the last certificate it encounters. If there is a chain this might be causing some unexpected behaviour?

colourful-land:
> Which XML file are you using? I have issues accessing Tokenscript.org. In any case, _very briefly_ looking at the code, it seems it will return the first public key it encounters, but the last certificate it encounters. If there is a chain this might be causing some unexpected behaviour?

@James-Sangalli please check all of the certificates and find at least 1 matching.

> have issues accessing Tokenscript.org. 

@jot2re it's split into two: https://community.tokenscript.org for the forum & http://tokenscript.org for the main website

jot2re:
> > Which XML file are you using? I have issues accessing Tokenscript.org. In any case, _very briefly_ looking at the code, it seems it will return the first public key it encounters, but the last certificate it encounters. If there is a chain this might be causing some unexpected behaviour?
> 
> @James-Sangalli please check all of the certificates and find at least 1 matching.
> 
> > have issues accessing Tokenscript.org.
> 
> @jot2re it's split into two: https://community.tokenscript.org for the forum & http://tokenscript.org for the main website

It seems the issue was that my browser would not try http://tokenscript.org when I simply wrote tokenscript.org. It would default to _only_ try https://tokenscript.org. However, after accessing http://tokenscript.org it defaults to this when I try tokenscript.org. Chrome seems to be able to find it though, so this might just be a Safari issue

James-Sangalli:
> > > @colourful-land @JamesSmartCell currently can check issuer against cacerts but noticed that it breaks when there is a DSA algorithm on signature.validate.
> > 
> > 
> > @jot2re I believe the algorithm is defaulting to RSA somewhere in here:
> > DOMValidateContext valContext = new DOMValidateContext(new CertKeySelector(), nl.item(0));
> > in CertKeySelector, could you point me in the right direction to set the algorithm properly using the property inside the XML file?
> 
> Which XML file are you using? I have issues accessing Tokenscript.org. In any case, _very briefly_ looking at the code, it seems it will return the first public key it encounters, but the last certificate it encounters. If there is a chain this might be causing some unexpected behaviour?

@jot2re that is unrelated, if I submit a certificate signed by an RSA key it will not have this exception. The moment I use one signed by a DSA key (DAI.tsml for example) it will throw this exception as the algorithm it is used for is RSA.

James-Sangalli:
Ok, it's working now but only for RSA...

colourful-land:
Nice

colourful-land:
Okay, my previous request of building a test case is still needed, but also have a look at the travis CI error

colourful-land:
@James-Sangalli I modified your test case (with 0 modification to your validator code) to demonstrate  ECDSA verification

lib/src/main/java/io/stormbird/token/tools/TokenScriptValidator.java
```diff
@@ -0,0 +1,219 @@
...
+                boolean refValid =
+                        ((Reference) i.next()).validate(valContext);
+                System.out.println("ref[" + j + "] validity status: " + refValid);
+            }
+        }
+
+        return result;
+    }
+
+    protected static X509CertImpl selectKeyData(List list)
```
James-Sangalli: @JamesSmartCell these names are very confusing for us.

colourful-land: > @JamesSmartCell these names are very confusing for us.

Before you blame anyone, think that not everybody gets your advantage of being born in an English speaking country.

Oh wait a minute, you are referring to James?

lib/src/main/java/io/stormbird/token/tools/TokenScriptValidator.java
```diff
@@ -0,0 +1,219 @@
...
+                {
+                    try
+                    {
+                        crt.checkValidity();
+                        if (recovered != null)
+                        {
+                            PublicKey certKey = crt.getPublicKey();
+                            if (Arrays.equals(recovered.getEncoded(), certKey.getEncoded()))
+                            {
+                                cert = crt;
```
James-Sangalli: @JamesSmartCell shouldn't it return once it finds the signing cert?

lib/src/main/java/io/stormbird/token/tools/TokenScriptValidator.java
James-Sangalli: Hmm? I clearly marked @JamesSmartCell not sure what you are referring to.

lib/src/main/java/io/stormbird/token/tools/TokenScriptValidator.java
```diff
@@ -0,0 +1,250 @@
...
+                            }
+                        }
+                        else
+                        {
+                            if (crt.getSigAlgName().equals("SHA256withECDSA"))
+                            {
+                                cert = crt;
```
James-Sangalli: Cannot recover the public key unless it is RSA? @JamesSmartCell @colourful-land

colourful-land: I do not know what the quoted code intends, but speaking from the point of Java Cryptography Architecture, it's clearly possible to get the EC public key from a cert. Just add this to see for yourself:

              169      try
              170      {
              171          crt.checkValidity();
              172          System.out.println(crt.getPublicKey().toString());

That outputs this:

    io.stormbird.token.tools.TokenScriptValidatorTest > verifyECDSAxmldsig STANDARD_OUT
    Sun EC public key, 256 bits
      public x coord: 24544810233457759196143703211090909005083847164792591927755581006921757564400
      public y coord: 53137840816744055503003735497360707636898773073824264522122612147305425927261
      parameters: secp256r1 [NIST P-256, X9.62 prime256v1] (1.2.840.10045.3.1.7)
    Sun EC public key, 256 bits
      public x coord: 24544810233457759196143703211090909005083847164792591927755581006921757564400
      public y coord: 53137840816744055503003735497360707636898773073824264522122612147305425927261
      parameters: secp256r1 [NIST P-256, X9.62 prime256v1] (1.2.840.10045.3.1.7)

(it outputs 2 times because I wasn't using your post-refactor code. @JamesSmartCell 's old code goes through certs 2 times)

James-Sangalli: This is making me more confused. Why recover the public key if it is going to checkValidity on the cert? My refactoring has made this function fail but I am having trouble pinpointing the issue. Still looking.

James-Sangalli:
@colourful-land please observe the TODO's I left in my latest commit

lib/src/main/java/io/stormbird/token/tools/TokenScriptValidator.java
```diff
@@ -101,19 +95,6 @@ private static TSMLValidationResult validateSSLCertificateIssuer(
                 }
             }
         }
-        else
-        {
-            result.isValid = false;
-            boolean sv = signature.getSignatureValue().validate(valContext);
-            System.out.println("signature validation status: " + sv);
-            // check the validation status of each Reference
-            Iterator i = signature.getSignedInfo().getReferences().iterator();
```
James-Sangalli: @colourful-land @JamesSmartCell why was this included? Doesn't change anything.

James-Sangalli:
@colourful-land solved the signature verification problem. Ready to merge (provided you are satisfied with the implementation).

lib/src/test/java/io/stormbird/token/tools/TokenScriptValidatorTest.java
```diff
@@ -0,0 +1,27 @@
...
+        assert(!result.isValid);
+    }
+
+    @Test
+    public void verifyECDSAxmldsig() throws Exception {
+        InputStream DAIToken = new FileInputStream("src/test/ts/DAI.tsml");
+        TSMLValidationResult result = TokenScriptValidator.validateXML(DAIToken);
+        assert(result.isValid);
+        assert(result.subjectPrincipal.equals("CN=*.aw.app"));
+    }
```
jot2re: I would stress test the solution a bit more by adding some negative tests, e.g. error in the certificate chain.

lib/src/main/java/io/stormbird/token/tools/TokenScriptValidator.java
```diff
@@ -0,0 +1,261 @@
...
+        // By default on Oracle JRE, algorithm is PKIX
+        TrustManagerFactory tmf = TrustManagerFactory
+                .getInstance(TrustManagerFactory.getDefaultAlgorithm());
+        // 'null' will initialise the tmf with the default CA certs installed
+        // with the JRE.
+        tmf.init((KeyStore) null);
+        X509TrustManager tm = (X509TrustManager) tmf.getTrustManagers()[0];
+        X509Certificate[] validIssuers = tm.getAcceptedIssuers();
+        int index = 0;
+        while(certList.size() > index)
```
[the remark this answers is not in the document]
colourful-land: Yes we should, @James-Sangalli There is not much performance punishment if you use an algorithm at O² efficiency so you can come up with something stupid.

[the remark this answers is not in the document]
n1v0lg: Hi, I believe you should be able to validate the cert chain using `CertPathValidator`: 

https://docs.oracle.com/javase/8/docs/api/java/security/cert/CertPathValidator.html#validate-java.security.cert.CertPath-java.security.cert.CertPathParameters-

Here is a code snippet of an alternative to the current `validateSSLCertificateBasedOnTrustedCA` method, that demonstrates the use of `CertPathValidator`:

https://gist.github.com/n1v0lg/4847b7821876412ebc5cf967070f8be3

I haven't been able to validate it works because the code currently fails with a Duplicate KeyValue error before it gets to this portion of the code.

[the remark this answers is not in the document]
colourful-land: > this approach requires that the certificate chain is correctly ordered, i.e., the certs are in reverse order of issuance (root CA comes last).

Is there a convention of such? If so, we can simply require them to be ordered in the documents of creation of such signatures.

James-Sangalli:
@colourful-land @jot2re @n1v0lg 
Fixed up the duplicate checker and added @n1v0lg chain verification code. I think it is safe to approve this PR provided that we all agree that the order for the chain certificate should be correct (Is this assumption valid, @n1v0lg?)

@n1v0lg for some reason, setting setRevocationEnabled to true causes the test to fail, the certificate is valid however and not revoked.

lib/src/main/java/io/stormbird/token/tools/TokenScriptValidator.java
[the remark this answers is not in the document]
James-Sangalli: Seems to attest here that it is supposed to be order like that but in practise many clients can verify it unordered. 

https://utcc.utoronto.ca/~cks/space/blog/tech/SSLChainOrder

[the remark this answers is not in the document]
n1v0lg: I think it would be ideal to order the certs as part of the validation, instead on relying that they are ordered. 

Basically, you just need to find the root cert, then the cert it signed, and so on by matching up `getSubjectDN` with `getIssuerDN` (see here for example implementation: https://www.programcreek.com/java-api-examples/?code=pinguo-sunjianfei/Android-Application-ZJB/Android-Application-ZJB-master/extras/zjb-sync-volley/src/main/java/com/zjb/volley/core/ssl/GenericX509TrustManager.java).

n1v0lg:
@James-Sangalli I figured it out I believe--to check cert revocation, it suffices to set:

`Security.setProperty("ocsp.enable", "true");`

With that property set, `setRevocationEnabled` can also be set to true (test passes fine). I've updated my gist accordingly.

**Note** This will use OCSP (online cert. status protocol) to verify that the cert is not revoked. This requires a request to an OCSP responder; if the responder is not reachable for any reason, this will fail and the certificate won't validate.

I second @jot2re that ideally there should be more tests for negative scenarios (e.g., untrusted intermediate CA, expired cert, revoked cert).

lib/src/main/java/io/stormbird/token/tools/TokenScriptValidator.java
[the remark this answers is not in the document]
James-Sangalli: @n1v0lg added the code, seems to be working and pretty straightforward with matching by subject and issuer.

colourful-land:
@James-Sangalli doesn't seem to work for me:

    $ curl -F 'file=@lib/src/test/ts/EntryToken.tsml' localhost:8080/api/checkSig
    {"result":"fail","failureReason":"Path does not chain with any of the trust anchors"}(Add-TokenScript-Validator-API)

If you think it's the problem of the file (`lib/src/test/ts/EntryToken.tsml`) then can you put a correct file in there? That's what other devs are supposed to try with, the test-case.

James-Sangalli:
> @James-Sangalli doesn't seem to work for me:
> 
> ```
> $ curl -F 'file=@lib/src/test/ts/EntryToken.tsml' localhost:8080/api/checkSig
> {"result":"fail","failureReason":"Path does not chain with any of the trust anchors"}(Add-TokenScript-Validator-API)
> ```
> 
> If you think it's the problem of the file (`lib/src/test/ts/EntryToken.tsml`) then can you put a correct file in there? That's what other devs are supposed to try with, the test-case.

As we have discussed before, EntryToken is signed by shong.wang which is signed by an invalid authority. Please try again with DAI.tsml for which there is also test case (EntryToken test is supposed to fail).

jot2re:
It looks good. But I have added a new issue #734 to address some desirable “hardening” aspects of the code.

Issue #690, opened 2019-05-11
Retrieve erc20 transactions via events from etherscan


James-Sangalli:
As seen in iOS here: https://github.com/AlphaWallet/alpha-wallet-ios/pull/1217

PR #691, opened 2019-05-11
start adding functionality to retrieve erc20 events from etherscan

James-Sangalli:
Intended for #690

Added the base links and a boolean check to distinguish but think it more efficient that you complete it to how you would like it to be (for consistency with the rest of the codebase).

It has been completed in iOS here: https://github.com/AlphaWallet/alpha-wallet-ios/pull/1217

Still needs to be configured with the appropriate calls and needs to be added to the NetworkInfo Object.


James-Sangalli:
@justindg would you be able to do this? basically this API allows you to pickup all erc20 transactions going to and from you. This enables better contract detection but it needs to be incorporated into the codebase to coincide with regular transaction retrieval and contract detection based on transactions.

James-Sangalli:
@JamesSmartCell says this is already done.

PR #700, opened 2019-05-22
adds an enable function which simply passes

James-Sangalli:
This PR solves for the case whereby the dapp browser asks to enable the web3 injection. Some sites do not catch for this and it can result in an unhandled exception. Since we have not created a privacy mode feature for our dapp browser yet, we should define it and allow it to pass. Carbon copy of https://github.com/AlphaWallet/alpha-wallet-ios/pull/1226


hboon:
@James-Sangalli you forgot to describe what the PR is for :P

James-Sangalli:
@JamesSmartCell don't forget to delete the branch! Yes, the dapp site asking for permission to access the web3 provider

Issue #513, opened 2019-01-24
Network page should be an activity not a dialog 


James-Sangalli:

JamesSmartCell:
It does seem like other wallets also implement this as an activity. Wonder why we designed it as a dialog in the first place or is it just a hangover from the inherited source?

JamesSmartCell:
Added @James-Sangalli to the assignee list so you may make the change.

colourful-land:
1. Why is such a need?
2. What network page? In dapp browser or in wallet?

James-Sangalli:
> 1. Why is such a need?
> 2. What network page? In dapp browser or in wallet?

Again, this was created before we had the dapp browser network setting.

It was intended to be for the settings page only, when you select the network.

PR #702, opened 2019-05-26
Convert Network Selection Dialog to Activity

justindg:
#513


JamesSmartCell:
Apologies for the long delay in checking.

The new activity does look better and means there's no need for scrolling to see all the options (at least on high res phones). Need to double check it scrolls correctly on a lower resolution phone. Would be great to have the selections colour coded like metamask:

networkselect

However the network selection on dapp browser should still be a dialog box, with a title 'Select Network'. The difference is that you're selecting a single network rather than choosing a filter. It can be an activity, but this doesn't make sense for the UX.

colourful-land:
@James-Sangalli 

1. Why is this approved when the ask is to make it into an activity and it ended up still a dialogue?

@justindg 

2. If this is about the pop up in dapp browser, Please don't show the networks not enabled in Settings, instead, tell the users to enable more network in Settings.

colourful-land:
@justindg 
it should 𝑛𝑜𝑡 be possible for the user to choose a network not previously enabled in Settings when the user browses a dapp, no matter how much convenience it brings, even if the browser can correctly identify the user should be accessing a network not previously enabled in Settings.

If you prioritize consitency, you get convenice. If you priorize convenience, you get neither consistency nor convenience.

colourful-land:
I would approve if this is done @JamesSmartCell 

1. Only list the enabled networks in the dialogue.
2. Hint that more networks are available in settings (yes, I know a message is not the best design)
3. The word "Ethereum" is made into a button, with a browser "refresh" icon followed by the text "Ethereum".

JamesSmartCell:
Just to reduce confusion: the spec @justindg was following was to convert all dialog into activity, and the previous operation of the network picker in dapp browser didn't restrict based on network filters.

Ok I have a design solution for the button; pending a redesign which fulfils the requirement. 

I'll update the PR.

James-Sangalli:
> @James-Sangalli
> 
> 1. Why is this approved when the ask is to make it into an activity and it ended up still a dialogue?
> 
> @justindg
> 
> 1. If this is about the pop up in dapp browser, Please don't show the networks not enabled in Settings, instead, tell the users to enable more network in Settings.

It is not a dialog, it is an activity.

JamesSmartCell:
It makes sense for the filters to be a separate activity but not for the network selector in dappbrowser.

colourful-land:
> It is not a dialog, it is an activity.

@James-Sangalli When you gave justing https://github.com/AlphaWallet/alpha-wallet-android/issues/513 do you mean:

𝑎) the network selector in Settings should be changed to an activity;
𝑏) the network selector in dapp should be changed to an activity;

I think Justin took your intention as 𝑎+𝑏 when you only intended 𝑎

I can't remember well but I must have lacked precision and was laconic too in my first start-up. It takes time to build that skill but I think the short cut is to speak verbosely at the outset and learn to make it terse later.

James-Sangalli:
> #513

The issue clearly says that it should be an activity, also, this issue was created long before network filters were made as it was done in Jan, before we had the multi chain UI.

The network selection page should still be an activity like this PR addressed.

JamesSmartCell:
With the UX overhaul it now makes sense for network select to be an activity. Will merge after testing.

colourful-land:
> The network selection page should still be an activity like this PR addressed.

Which is the selection page?

𝑎) the network filter in Settings 
𝑏) the network selector in dapp

James-Sangalli:
@JamesSmartCell Please remember to delete the branch and close the issue

PR #706, opened 2019-05-31
Fix for google search not working on Android 9

JamesSmartCell:
Need to add this line to the manifest in order to allow google search in the dapp browser on Android 9.


hboon:
@JamesSmartCell Is it because the google URL constructed for searching is `http://`?

JamesSmartCell:
Yes that's right, somehow the google URL had been added as http: not https:

PR #711, opened 2019-06-10
Release 1.60.4

JamesSmartCell:
Release 1.60.4


James-Sangalli:
@JamesSmartCell no need to do this, simply mark a release on the front panel and it will create a tag.

James-Sangalli:
@JamesSmartCell we are also on version 2 since TokenScript is a big change. Please up it to version 2 and rebase.

PR #714, opened 2019-06-13
Blockscout ERC20

JamesSmartCell:
**Integrate Blockscout queries into the wallet view**

Blockscout fetches all ERC20 token balances (on a single chain) in a single API call, which means the wallet will discover all ERC20 token balances very quickly, and at the same time saving a lot of individual Infura queries.

The Infura query system is still operating alongside the Blockscout fetches.

The heuristic works as follows:

- each token records when it was last updated.
- each token is given an 'update weighting' based on importance - base currencies receive a high weighting, any chain with actual value also receives a high weighting, so the base currencies with actual value have the highest weighting, followed by tokens on those chains followed by testnet base currencies followed by tokens on testnets.
- every 500ms the wallet determines which token has the highest update pressure - ie a combination of when the token/currency was last checked multiplied by the update weighting. The token also cannot have been checked in the last 20 seconds (unless there's been a manual refresh).

Because the blockscout queries update the token check time for affected tokens it means most of the time ERC20's don't need to be balance checked, unless blockscout is offline.

This PR enables rapid discovery of new tokens (eg KNC, SAUD etc) so most users will never need to go into the 'Add Token' dialog again; ERC20 will appear in their wallet when it comes into the account, creating an improved user experience (wallet works as user would expect).

It also picks up airdrop tokens (this may or may not be a good thing, but it's still fun to get air drops appearing).

It also reduces the amount of network usage while blockscout is functional.


James-Sangalli:
@JamesSmartCell replaces https://github.com/AlphaWallet/alpha-wallet-android/pull/691?

JamesSmartCell:
This error is unrelated to the PR. I'm working on it now.

It doesn't use events, so no I would say it doesn't replace #691.

Which version of the OS on the emulator are you using, can you specify?

JamesSmartCell:
Crash should be solved by PR #716

PR #719, opened 2019-06-17
Dmz issue patch

James-Sangalli:
This PR patches the issues we had with the DMZ server serving pages relating to magic links. 

Changes:

- Only show a web page if it has a tsml file, else return a 404 (I think this makes sense because we do not want to show unverified token links nor should we say ones without tsml as 'unavailable' as this gives a false impression).
- Remove currency link support, this was buggy and a feature which is not very useful to begin with (MakerDAO team bailed on it). It also requires paymaster support to prevent front running.

I keep this as WIP until we can agree these changes are correct or suggest a different path. For the first point, it may simply be more elegant to show a page that says the token is available but has no corresponding tsml file (that would be a separate PR however).


hboon:
> [1. ]Only show a web page if it has a tsml file, else return a 404 (I think this makes sense because we do not want to show unverified token links nor should we say ones without tsml as 'unavailable' as this gives a false impression).

>  For the first point, it may simply be more elegant to show a page that says the token is available but has no corresponding tsml file (that would be a separate PR however).

Unless there are constraints, I think we should support showing the token as available (or not available) even if there is no TSML file (as a separate PR). It's at least useful during development when the TSML file hasn't been deployed on the repo server yet.

How does the rendering for the token card view work on the web page currently? Does it support function-origin based attributes?

James-Sangalli:
> > [1. ]Only show a web page if it has a tsml file, else return a 404 (I think this makes sense because we do not want to show unverified token links nor should we say ones without tsml as 'unavailable' as this gives a false impression).
> 
> > For the first point, it may simply be more elegant to show a page that says the token is available but has no corresponding tsml file (that would be a separate PR however).
> 
> Unless there are constraints, I think we should support showing the token as available (or not available) even if there is no TSML file (as a separate PR). It's at least useful during development when the TSML file hasn't been deployed on the repo server yet.
> 


@hboon that's what the local environment is for? When you test you simply do so locally with the tsml file and then later deploy to the server.

hboon:
>> Unless there are constraints, I think we should support showing the token as available (or not available) even if there is no TSML file (as a separate PR). It's at least useful during development when the TSML file hasn't been deployed on the repo server yet.

> @hboon that's what the local environment is for? When you test you simply do so locally with the tsml file and then later deploy to the server.

I was thinking of something like this during development, or a hackathon:

1. Deploy a contract
2. Generate MagicLink
3. Maybe write TokenScript. But haven't deployed to repo server yet. This takes time or is unnecessary. We can send the file directly to A via email/iOS AirDrop.
4. Send MagicLink to A
5a When A taps the link without the app installed (or on a desktop), the webpage shows an error.
5b AlphaWallet app is installed, A taps the link and imports it properly

It isn't critical, but sounds like an arbitrary omission to not support 5a when 5b is supported. Is there a reason to not support it?

James-Sangalli:
@hboon fair.

JamesSmartCell:
The recent patch enables support of ERC875 tokens without a definition. Useful for people doing experiments with tokens.

PR #730, opened 2019-06-30
ERC721 Tokenscript

JamesSmartCell:
Upgrade tokenscript handling to be ERC721 compatible.

This was done to make tokenscript handling for all tokens more generic - previously ERC721 wasn't supported. The upgrade allowed the Cryptokitties Tokenscript to be demoed at the Ethereum conference; 

Once we have the kitties Tokenscript working on iOS then we can release it to the server. The script may require a little formatting to make it look a bit better.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/entity/Ticket.java
+++ b/app/src/main/java/io/stormbird/wallet/entity/Ticket.java
@@ -1,48 +1,36 @@
 package io.stormbird.wallet.entity;
 
 import android.content.Context;
-import android.content.Intent;
 import android.os.Parcel;
 import android.os.Parcelable;
-import android.support.v7.widget.AppCompatRadioButton;
 import android.util.Log;
 import android.view.View;
 import android.widget.LinearLayout;
 import android.widget.ProgressBar;
 import android.widget.TextView;
-
-import io.reactivex.android.schedulers.AndroidSchedulers;
-import io.reactivex.schedulers.Schedulers;
-import io.stormbird.token.entity.*;
-import io.stormbird.token.tools.TokenDefinition;
-import io.stormbird.token.util.DateTime;
-import io.stormbird.token.util.DateTimeFactory;
-
-import io.stormbird.wallet.C;
-import io.stormbird.wallet.entity.opensea.Asset;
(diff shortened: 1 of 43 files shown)


James-Sangalli:
@JamesSmartCell iOS is already able to get the token id from the tokenscript which is passed through via the opensea call. I am not quite sure what this PR is supposed to mean because it is already possible to do so with the current implementation as when you have the token id you can get all that is required to display a token like cryptokitties. 

Please correct my misunderstanding if there is one.

JamesSmartCell:
Ah yes, the PR includes the multiple tokenId handling patch. That should have been in the description above...

Currently iOS won't handle kitties breeding script as it requires multiple tokenId handling. Discussed with Boon while in China he agrees with the approach as detailed in the tokenscript forum 'handling multiple tokenIds'. It's a draft application currently, pending comments.

Previously Android didn't completely handle tokenscript in ERC721. This oversight has been corrected in this PR and all common handling with ERC875 has been abstracted as per OO methodology (eg move common functionality upwards in the class structure as far as possible). The comment on the commit is 'Abstract' as a verb not because the commit is 'abstract'. It could be that this commit message was a bit too abstract?

I still haven't fully adjusted to working with squash yet - it's a battle between getting the branch merge-ready and using squash (using git merge master usually kills any option of a clean squash). Still working on it... I don't like the alternative of waiting until ready for PR then one big atomic commit.

hboon:
Just want to raise @colourful-land's attention to 1 bit: this will require a change to TokenScript itself. I remember you have some ideas and maybe plans about `<ts:selections>`. This probably affects them.

JamesSmartCell:
Thanks @hboon , yes this should also be included into the comment thread on this topic in TS development here: 
https://community.tokenscript.org/t/using-tokenid-as-input-to-transaction/237

James-Sangalli:
@JamesSmartCell rebased, hope you wont mind, kept a copy incase.

JamesSmartCell:
No worries, maybe you can give me another lesson on rebase ...

James-Sangalli:
@JamesSmartCell why is this closed?

PR #733, opened 2019-07-04
Fix auto download

JamesSmartCell:
Fix for auto download file location (for Chinese users mainly), plus fix the resource leak observed from time to time if the server can't find the location.


colourful-land:
Hi. The CCP uses something called random forest and it might hold the connection for an undetermined amount of time even if the apk is not in blacklist if the user wins some lottery. Can you tell me whether or not that piece of code would freeze the app should that happen?

JamesSmartCell:
2000 ms timeout for establishing connection.

James-Sangalli:
@colourful-land on you to merge this PR once your concerns are met.

JamesSmartCell:
Also, the thread runs in the background, so it doesn't hold up wallet startup.

Issue #731, opened 2019-07-03
Replace icons in the Tab Bar


tomekalphawallet:
Image resources can be found:

https://app.zeplin.io/project/5d088205bff2d15de6a4397b/dashboard?seid=5d11e61ebcecab70f88959bf

All 4 icons are to be replaced and the text follows the icons (explicit text).

tomekalphawallet:
iOS implementation can be found here:
https://github.com/AlphaWallet/alpha-wallet-ios/issues/1273

tomekalphawallet:
Just a comment on this one. We can improve that before the next release. 

Video comment:
https://www.dropbox.com/s/m7uuvndzu8x7zuf/09.06-tab-bar.mov?dl=0

- The active tab label text should be blue #1899c2
- Increase touch target size to labels too

PR #736, opened 2019-07-05
add new toolbar icons with text

James-Sangalli:
For #731


JamesSmartCell:
Nice work, looks a lot better than before.

Just two things: first is that the icon is now too close to the top of the border, it doesn't quite look right. easiest fix is to put the background white into the holding linearlayout view and add something like android:paddingTop="4dp" (or however much works) to that, or you can add the same padding to each of the 4 linearlayout icon/text holder views.

The other is the strings addition. In order to make it easy for translation when new strings are added, make sure to add the same string to the cn and es strings.xml. You can do that automatically through Intellij in the dialog popup when you ctrl+enter:

addstring

or manually whichever you like.

James-Sangalli:
@JamesSmartCell done, padded the top with the highest hierarchy linearlayout and added the transaction strings for Spanish and Chinese

PR #737, opened 2019-07-05
adding usage notes for XMLDSig verifier API

colourful-land:


hboon:
Can I summarize it as:

```
let results = makeWebApiCall()
if results.result == "success" {
    let domain = results.subject.split("=", maxSeparatorOnceOnly)[1] //just split once
    //good and we display the value of domain
} else {
    //failed
}
```

James-Sangalli:
> @colourful-land Can I summarize it as:
> 
> ```
> let results = makeWebApiCall()
> if results.result == "success" {
>     let domain = results.subject.split("=", maxSeparatorOnceOnly)[1] //just split once
>     //good and we display the value of domain
> } else {
>     //failed
> }
> ```
> 
> ?

@hboon yes, it will give back a result of either success or fail. If failed it will provide a failureReason and if success, it will provide information about the issuer

hboon:
@James-Sangalli thanks!

PR #740, opened 2019-07-06
Fixes from crashlytics

JamesSmartCell:
The following Null Pointer crashes fixed from Crashlytics reporting:

AssetDefinitionService: being passed a null contract address from wallet token list, reported from a fork of Alphawallet. Not sure what's going on in the modified code but the issue can be patched to ensure it doesn't cause a problem for us.

AddTokenActivity: Switching to an unavailable network - possibly a user build since only available networks are added to the network selector.

DappBrowserFragment QR handling warning: Crash when context switched out. This could occur when a user task switches while the camera is open, causing the QR scanner to return null, and the app tries to display the 'invalid QR' message but there's no context available.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,7 +38,7 @@ android {
         applicationId "io.stormbird.wallet"
         minSdkVersion 18
         targetSdkVersion 28
-        versionCode 58
+        versionCode 59
         versionName "2.00.3"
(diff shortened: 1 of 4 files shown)


JamesSmartCell:
Yeah that internal version change sneaks in - without it the app won't install even in debug mode as it's protecting me from installing an older app.

PR #742, opened 2019-07-07
adds user experior to help with UX

James-Sangalli:
Completes https://github.com/AlphaWallet/web-service/issues/46


JamesSmartCell:
Need to ensure users have consented to gestures being tracked. Should add a consent screen into the settings where users can consent to this.

PR #743, opened 2019-07-08
Fix swipe event issue

JamesSmartCell:
As reported by @zhangzhongnan928 on the Telegram group. 

It was impossible to scroll down on a website that was slightly longer than the screen height as this gesture was intercepted as a screen refresh event.

This PR makes the swipe refresh event detection less sensitive; specifically to trigger it you need to swipe downward more than 400 pixels within 300ms. These figures were arrived at by collecting swipe gesture pattern data; typically a swipe refresh fling gesture lasts only 100ms and will be well over 500 pixels. Using the more generous data still allows easy refresh but lets through browser scrolling.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/ui/DappBrowserFragment.java
+++ b/app/src/main/java/io/stormbird/wallet/ui/DappBrowserFragment.java
@@ -28,6 +28,8 @@ import android.widget.*;
 
 import com.google.gson.Gson;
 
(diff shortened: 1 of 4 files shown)


zhangzhongnan928:
I don't think the issue is from "swipe refresh event detection"

Uniswap on my Andriod is twice as long as the screen, it is not "a website that was slightly longer than the screen height"

The screen is not moving at all when scrolling down.

PR #744, opened 2019-07-08
Fix dapp browser refresh issue

JamesSmartCell:
Fixing a double issue with DappBrowser usability. 

@zhangzhongnan928 reported an issue on Telegram where on some aspect ratios the Uniswap Dapp page couldn't be scrolled down. This is due to the web page layout - a header locks the outer page scroll to be always at the top, while an inner view scrolls up/down. This confuses the native refresh pull down system, which doesn't pass the down gesture on to the underlying page. I overrode with a simple algorithm to make the pull down refresh less sensitive.

A related issue can be observed in the 'Cryptocare' dapp. On the page there are left/right sliders and clicking on the 'causes' link there are causes you can browse by scrolling left and right.

The left/right scroll gestures were intercepted by the viewpager layer and just go to shift the page to transactions or settings. Implementing a simple algorithm to measure the velocity and fling distance; if velocity is over a certain threshold and the fling distance is more than 60% then it's passed to view pager to change page. Anything less is passed through into the dapp browser screen.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/ui/DappBrowserFragment.java
+++ b/app/src/main/java/io/stormbird/wallet/ui/DappBrowserFragment.java
@@ -943,6 +943,17 @@ public class DappBrowserFragment extends Fragment implements
     @Override
     public void RefreshEvent()
     {
-        web3.reload();
+        //determine scroll position
+        Log.i("Touch", "SCROLL: " + web3.getScrollY());
+        if (web3.getScrollY() == 0)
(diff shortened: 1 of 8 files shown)


JamesSmartCell:
@hboon yes there is a standard refresh button - it's buried in the hamburger menu. You can also just click on the 'change network' it will also refresh. I've been using the new system all day in conjunction with debugging another issue and it feels like it's working well.

Issue #745, opened 2019-07-08
"Invalid Transaction" Error When Sending Simple Tx Through DApp


nickreynolds:
sending a simple transaction via the web3 call `web3.sendTransaction({from:"0x123...", to: "0x456...", value "10000"})` pops the "Invalid Transaction" error modal (which cannot be closed). Same function call correctly allows transaction to be sent using MetaMask (extension or mobile browser beta).

Using web3 version 0.20.3

Would be nice if the error dialog presented more helpful information about why the transaction failed.

Would also be nice if the app supported webview debugging so I could connect to Chrome Devtools to check logs.

PR #746, opened 2019-07-09
Fix standard eth transaction

JamesSmartCell:
Addressing #745, standard ETH dapp transactions have never been supported.

Logic was changed to ensure all types of Dapp transaction are handled correctly, and also added a 'check contract' button in the confirmation screen so the user can check the contract they're calling on etherscan.


James-Sangalli:
@JamesSmartCell you are saying that web3.eth.sendTransaction({value: xx}) or any contract call that had a value param doesn't work in Android until this PR?

JamesSmartCell:
If it was a transaction that had no payload (ie pure eth send) it wouldn't previously have worked. A transaction with value and payload would have passed ok.

PR #752, opened 2019-07-11
Harden validator api

### LINKED ISSUE #734: Harden pull request #683
Since pull request #683 considers security critical checks it would be good to ensure full documentation along with lots of negative tests. For example the following. 

- [x] Add method signatures including which security critical the given method checks for.
- [x] Add more negative test: by including certs in the chain that are not yet valid and that has expired. If possible also try to include a revoked certificate. 
- [x] Try to stress the implementation by having the certificate chain in an unexpected order. 
- [x] Try to stress the implementation by having more than expected KeyInfo (and others) elements in the xml files. 
- [x] Ensure that there are positive tests for both RSA and ECDSA.

James-Sangalli:
Closes #734


James-Sangalli:
@colourful-land @JamesSmartCell @jot2re ready to merge now.

PR #784, opened 2019-08-08
added schema support for july

James-Sangalli:


James-Sangalli:
@JamesSmartCell bump

app/src/main/java/io/stormbird/wallet/entity/tokenscript/TokenscriptFunction.java
```diff
@@ -67,6 +274,74 @@ public Function generateTransactionFunction(String walletAddr, BigInteger tokenI
                             break;
                     }
                     break;
+                case "string":
+                    params.add(new Utf8String(arg.element.value));
```
JamesSmartCell: Should there be a `break;` statement after each of these `case:` switches?

app/src/main/java/io/stormbird/wallet/entity/tokenscript/TokenscriptFunction.java
```diff
@@ -43,6 +47,209 @@ public Function generateTransactionFunction(String walletAddr, BigInteger tokenI
             resolveReference(arg, tokenId, definition, attrIf);
             switch (arg.parameterType)
             {
+                case "e18":
```
JamesSmartCell: I don't think e18 should go in here - it should be handled before we get here probably. It should be handled more programmatically anyway (my fault probably) that is the number should be read and the input scaled accordingly.

colourful-land: It seems sangalli has moved on, can you fix this after him?

hboon: > that is the number should be read and the input scaled accordingly.

If we decide to do this, can we write it in the doc and create an issue on iOS to do that too? Right now, "e18" is considered as as special case, we don't handle arbitrary "eN", where N is a whole number.

James-Sangalli: @hboon @colourful-land we should handle eN because e18 is not the only possibility

colourful-land:
I think this one can wait for the namespace change to ease JB's work on changing namespace

app/src/main/java/io/stormbird/wallet/entity/tokenscript/TokenscriptFunction.java
colourful-land: I proposed before and I do it again: I propose only implement a few (no more than 4) types of e, like e8, e18, e10 etc.

The reason being if you want to generalise it you may as well generalise it properly, e.g. why not raise the number to base 2? Let's say we use b24 to represent that the value has to be bit-shifted 24 positions.

James-Sangalli: I can put in some popular e multipliers into the new schema and change this as we move to october schema.

PR #785, opened 2019-08-08
Implement hd keys

colourful-land:
- MinSDK is now 23
- Compile android level is now 29.

James will come up with his side of comment. I'm doing PR so Nikolaj can comment on key mgmt


app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -0,0 +1,1412 @@
...
+        // and then we can return your byte array.
+        return byteBuffer.toByteArray();
+    }
+
+
+    private synchronized void deleteKey(KeyStore keyStore, String keyAddr)
+    {
+        try
+        {
+            String encryptedBytesPath = getEncryptedFilePath(keyAddr);
```
n1v0lg: If deleting files causes an exception, the flow never reaches key deletion from keystore. Make all deletions not contingent on each other, i.e., delete as much as possible rather than "aborting" if a preceding delete operation fails.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -0,0 +1,1412 @@
...
+        if (nonAuthKey.exists()) nonAuthKey.delete();
+        if (nonAuthKeystore.exists()) nonAuthKeystore.delete();
+        if (legacyKey.exists()) legacyKey.delete();
+    }
+
+    /**
+     * Delete all traces of the key in Android keystore, encrypted bytes and iv file in private data area
+     * @param keyAddress
+     */
+    public synchronized void deleteKeyCompletely(String keyAddress)
```
n1v0lg: Same high level comment as for `deleteKey`. Delete as much as possible rather than "aborting" if a preceding delete operation fails.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -0,0 +1,1412 @@
...
+                return false;
+            case CREATE_KEYSTORE_KEY:
+            case CREATE_PRIVATE_KEY:
+                return deviceIsLocked();
+            default:
+                return true;
+        }
+    }
+
+    private String getEncryptedFilePath(String keyAddr)
```
n1v0lg: It would be more robust to explicitly specify what type of file (i.e., auth-locked, non-auth-locked, etc) we are trying to fetch, and fail if it isn't there. Otherwise, we can run into some bad flows. For example, suppose the attacker has access to the file system. He could then remove an existing HD-auth-locked seed file, and copy back on old HD-non-auth-locked seed file; this method would then silently return the HD-non-auth-locked seed file to the caller, even though the caller expects an HD-auth-locked seed file. 

An even more robust approach would be to tie the key path names directly to the secure enclave, i.e., if the AES key in the enclave is not auth-locked, use "...-no-auth-hd", else use "...-hd".

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -0,0 +1,1412 @@
...
+    private void importHDKey()
+    {
+        //first recover the seed phrase from non-authlocked key. This removes the need to keep the seed phrase as a member on the heap
+        // - making the key operation more secure
+        String seedPhrase = unpackMnemonic(currentKeyAddress, UPGRADE_HD_KEY);
+        HDWallet newWallet = new HDWallet(seedPhrase, "");
+        storeHDKey(newWallet, Operation.IMPORT_HD_KEY);
+    }
+
+    private UpgradeKeyResult upgradeKey()
```
n1v0lg: Assuming the attacker has write access to the file system, there is a potential (non-severe) attack here: assume we have a non-auth-locked seed for address ADDR. The attacker can create a bogus file named ADDR-hd (note no no-auth label); in this case the if branch doesn't run, and instead the user is given to believe that the seed is already auth-locked. 

The first problem here is how `getEncryptedFilePath` is implemented (see that method for comments); the bigger problem is that file names are used as a source of truth. It would be better to tie checking if a key is auth-locked to the secure enclave instead.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -0,0 +1,1412 @@
...
+                iv = readBytesFromFile(getFilePath(context, keyAddr + "iv"));
+            if (iv == null || iv.length == 0)
+            {
+                keyFailure("Cannot setup wallet seed.");
+                return "";
+            }
+            Cipher outCipher = Cipher.getInstance(CIPHER_ALGORITHM);
+            final GCMParameterSpec spec = new GCMParameterSpec(128, iv);
+            outCipher.init(Cipher.DECRYPT_MODE, secretKey, spec);
+            CipherInputStream cipherInputStream = new CipherInputStream(encryptedHDKeyBytes, outCipher);
```
n1v0lg: If tag verification fails during decryption you simply print a stack trace. Presumably we want to do something else since this would mean that the stored key is destroyed (and possibly maliciously tampered with) and we would want to start a restore flow/notify the user.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -0,0 +1,1412 @@
...
+    private void createHDKey()
+    {
+        HDWallet newWallet = new HDWallet(DEFAULT_KEY_STRENGTH, "");
+        storeHDKey(newWallet, Operation.CREATE_NON_AUTHENTICATED_KEY); //create non-authenticated key initially
+    }
+
+    /**
+     * Called after an authentication event was required, and the user has completed the authentication event
+     */
+    private void importHDKey()
```
n1v0lg: A high level comment about the import flow. The current flow (1) restores the seed from non-auth-locked key `k0`; (2) creates auth-locked key `k1` and attempts to store the seed encrypted under that; in case the user is not authenticated, we (3) create non-auth-locked key `k2`, store seed encrypted under that, request authentication and then basically repeat the flow but with `k2` as the new auth-locked key.

Step 3 seems redundant: couldn't we instead first check if the user is authenticated, and only then create the auth-locked key `k1` (overwriting `k0` which is now safe to do)? If the user fails to authenticate, we simply issue a warning and don't to anything else.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -0,0 +1,1412 @@
...
+
+    private synchronized void storeHDKey(HDWallet newWallet, Operation operation)
+    {
+        PrivateKey pk = newWallet.getKeyForCoin(CoinType.ETHEREUM);
+        currentKeyAddress = CoinType.ETHEREUM.deriveAddress(pk);
+
+        storeEncryptedBytes(currentKeyAddress, newWallet.mnemonic().getBytes(), operation);
+    }
+
+    private synchronized void storeEncryptedBytes(String address, byte[] data, Operation operation)
```
n1v0lg: There are a lot of different flows that pass through here and it's difficult to review each possible one and ensure that, for example, keys and files are always deleted when necessary. I understand that this is done to limit code duplication, but IMO the code is too complex.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -0,0 +1,1412 @@
...
+        wallet.checkWalletType(context);
+        switch (wallet.type)
+        {
+            case KEYSTORE_LEGACY:
+                return signWithKeystore(wallet, transactionBytes);
+            case KEYSTORE:
+                return signWithKeystore(wallet, transactionBytes);
+            case HDKEY:
+                String mnemonic = unpackMnemonic(currentKeyAddress, Operation.SIGN_DATA);
+                if (mnemonic.length() == 0)
```
n1v0lg: If mnemonic is length 0, a 4 byte array is returned which is used in MarketQueueService.sigFromByteData. In particular position 64 in this array is accessed which should throw an ArrayIndexOutOfBoundsException. Is this properly handled somewhere?

app/src/main/java/io/stormbird/wallet/service/KeystoreAccountService.java
```diff
@@ -0,0 +1,408 @@
...
+                wallet.type = WalletType.KEYSTORE;
+                wallets.add(wallet);
+            }
+
+            return wallets.toArray(new Wallet[0]);
+        })
+        .subscribeOn(Schedulers.io());
+    }
+
+    private byte[] patchSignatureVComponent(byte[] signature)
```
n1v0lg: Could you document this method? A little unclear what the byte manipulations below achieve.

app/src/main/java/io/stormbird/wallet/service/KeystoreAccountService.java
```diff
@@ -0,0 +1,408 @@
...
+    public Single<String> exportAccount(Wallet wallet, String password, String newPassword) {
+        return Single
+                .fromCallable(() -> getCredentials(keyFolder, wallet.address, password))
+                .map(credentials -> org.web3j.crypto.Wallet.createLight(newPassword, credentials.getEcKeyPair()))
+                .map(objectMapper::writeValueAsString)
+                .subscribeOn(Schedulers.io());
+    }
+
+    /**
+     * TODO: Delete file, check if corresponding key is deleted
```
n1v0lg: Fix TODO.

app/src/main/java/io/stormbird/wallet/service/KeystoreAccountService.java
```diff
@@ -0,0 +1,408 @@
...
+        try {
+            JSONObject jsonObject = new JSONObject(store);
+            return "0x" + Numeric.cleanHexPrefix(jsonObject.getString("address"));
+        } catch (JSONException ex) {
+            throw new Exception("Invalid keystore");
+        }
+    }
+
+    /**
+     * TODO: Import private key to keystore
```
n1v0lg: Fix TODO.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -399,15 +394,14 @@ private void getAuthenticationForSignature()
...
-            if (!fileExists || secretKey == null)
+            SecretKey secretKey = (SecretKey) keyStore.getKey(currentWallet.address, null);
+            String encryptedHDKeyPath = getFilePath(context, currentWallet.address);
+            if (!new File(encryptedHDKeyPath).exists() || secretKey == null)
             {
                 signCallback.GotAuthorisation(false);
                 return;
             }
-            byte[] iv = readBytesFromFile(getFilePath(context, currentKeyAddress + "iv"));
+            byte[] iv = readBytesFromFile(getFilePath(context, currentWallet.address + "iv"));
```
jot2re: There might still be a slight issue by using the address of a wallet as a file identifier as it could allow someone with renaming access to the file system to switch around two wallets. Of course transactions would only go through if the same mnemonic was used for both wallets. Thus this seems to be a very minor issue.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -528,99 +504,91 @@ private void importHDKey()
     {
         //first recover the seed phrase from non-authlocked key. This removes the need to keep the seed phrase as a member on the heap
```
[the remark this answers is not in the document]
JamesSmartCell: I didn't know that. I had thought that stack/heap worked the same in Java as in C++ but I suppose there's no reason why it would. Doing it this way does make the code slightly cleaner and ensures if something goes wrong while getting authentication the key will still be there.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -465,25 +458,13 @@ private synchronized String unpackMnemonic(String keyAddr, Operation operation)
...
-                    break;
-            }
+            return new String(mnemonicBytes);
         }
         catch (InvalidKeyException e)
         {
             if (e instanceof UserNotAuthenticatedException)
             {
-                checkAuthentication(operation);
+                return REQUIRE_AUTHENTICATION;
```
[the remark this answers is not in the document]
JamesSmartCell: Agree. Have refactored.

app/src/main/java/io/stormbird/wallet/service/KeyService.java
```diff
@@ -434,27 +428,26 @@ private void getAuthenticationForSignature()
         signCallback.GotAuthorisation(false);
     }
 
-    private synchronized String unpackMnemonic(String keyAddr, Operation operation)
+    private synchronized String unpackMnemonic()
```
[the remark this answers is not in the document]
JamesSmartCell: Agree. Have refactored.

JamesSmartCell:
If there's no other issues I will approve this branch then merge with master

colourful-land:
> If there's no other issues I will approve this branch then merge with master

Not without solving the conflicts?

PR #806, opened 2019-08-14
Using Lambda to get Express-Of-Trust address

colourful-land:
To try this:

1. Build apk (you need to `make clean` first if you have been working on `Implement-HD...` branch)
2. Run the new build and go to Settings -> Help & FAQ
3. Click the blank region between "How do I transfer ETH into my wallet" and "Sett need help? Contact Us"
4. Wait for 6s (sorry, will improve this)

You should see a toast message which gives the express-of-trust address generated from a pre-configured tokenscript and contract address. The value should be the same found in the test case `lib/src/test/java/io/stormbird/token/tools/TrustAddressGeneratorTest.java`


app/src/main/java/io/stormbird/wallet/ui/TrustAddressGenerator.java
```diff
@@ -0,0 +1,23 @@
+package io.stormbird.wallet.ui;
+
+import com.amazonaws.mobileconnectors.lambdainvoker.LambdaFunction;
+
+public interface TrustAddressGenerator {
+
+    /**
+     * Invoke the Lambda function "TokenScriptTrustAddress".
+     *
+     * Change the method name in this Interface to fit the
```
James-Sangalli: This belongs in the ui folder?

colourful-land: Will change it.

PR #817, opened 2019-08-18
Add required types for tokenscript

### LINKED ISSUE #814: Transactions involving ether crash
Go to WETH tokencard and wrap ether for WETH, it will crash. We need to fix this before the tutorial

JamesSmartCell:
Fix for tokenscript native send function (determine correct send address) and add tokens which are defined in debug tokenscript to tokens list.

Fixes #814

### DIFF
--- a/lib/src/main/java/io/stormbird/token/tools/TokenDefinition.java
+++ b/lib/src/main/java/io/stormbird/token/tools/TokenDefinition.java
@@ -434,23 +434,23 @@ public class TokenDefinition {
                             if (thisDate.equals(schemaDate))
                             {
                                 //all good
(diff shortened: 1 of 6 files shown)


lib/src/main/java/io/stormbird/token/tools/TokenDefinition.java
```diff
@@ -29,7 +32,7 @@
     public TokenscriptContext context;
     public String holdingToken;
 
-    public static final String TOKENSCRIPT_CURRENT_SCHEMA = "2019/05";
+    public static final String TOKENSCRIPT_CURRENT_SCHEMA = "2019/07";
```
James-Sangalli: @JamesSmartCell should still be 05

PR #837, opened 2019-08-28
add lambda for apple association file

James-Sangalli:
Working on AWS CN


hboon:
It'll work if routing is done correctly, but note that this is serving static content with Lambda.

James-Sangalli:
@hboon @colourful-land 

Actually, no need for this. It's static so it shouldn't use lambda and do a cold start for such a simple return.

PR #838, opened 2019-08-28
remove redundant agent from api

James-Sangalli:


hboon:
> @hboon I dimly remembered user-agent being used in repo server but this is not the repo server so I'm approving this.

Right. Only the repo server captures it because we specify it in the spec, but I don't think we use it (yet).

PR #842, opened 2019-09-08
Token script certificate check

JamesSmartCell:
Check Tokenscripts using the API call to /v1/verifyXMLDSig.
Allow user to click on XML and TSML files in chat apps/email to import directly into AlphaWallet.


JamesSmartCell:
@james-sangalli it's not possible from the libs. Google removed the required certificate handling libraries from the Android security libs. Otherwise, yes it would be better to use the library.

JamesSmartCell:
Ok great - the actual security check is a small part of the PR and can easily be changed when the new code is ready.

PR #853, opened 2019-09-11
Fix for sendUncheckedTransaction freeze issue

### LINKED ISSUE #846: EtherJS signer.sendUncheckedTransaction freezes with Alpha Wallet on Android
https://github.com/ethers-io/ethers.js/issues/603
@ethereumgeek

JamesSmartCell:
Fixes #846. Callback was using incorrect return type and not actually passing any data.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/C.java
+++ b/app/src/main/java/io/stormbird/wallet/C.java
@@ -7,6 +7,7 @@ public abstract class C {
     public static final int SHARE_REQUEST_CODE = 1003;
     public static final int REQUEST_SELECT_NETWORK = 1010;
     public static final int REQUEST_BACKUP_WALLET = 1011;
+    public static final int REQUEST_TRANSACTION_CALLBACK = 1012;
 
     public static final String ETHEREUM_NETWORK_NAME = "Ethereum";
     public static final String CLASSIC_NETWORK_NAME = "Ethereum Classic";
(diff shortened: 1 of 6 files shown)


hboon:
@JamesSmartCell is this something that might affect iOS?

JamesSmartCell:
@hboon, yes it does but you may have already implemented it correctly. When you send a transaction formatted from a dapp you'll first go to the confirmation page to confirm. On return from the confirmation page you need to complete the callback in JS - that is in ```init.js```:

```
executeCallback(callbackId, null, hash)
```

You can confirm if iOS is working correctly by using the page provided here: https://ethersjs.web.app

Set dapp browser to rinkeby and click on the ```clicking here``` hyper link. After return from confirmation page the ```not started``` text should change to ```has hash``` then after it's written to a block it'll change to ```has receipt```. If you don't return the correct type in the callback you'll see ```has error``` and if you don't return the callback you'll see ```prompt user```. 


Note that for a transaction return the last param is the transaction hash, not the signature.

hboon:
@JamesSmartCell do you have screenshots of what should happen? I tried MetaMask, Trust wallet, Opera Touch and they all failed the test.

JamesSmartCell:
I think they threw this sample together for us to test, I had to reload a few times, keep clicking on the 'clicking here' hyperlink until you get the confirmation page then you should be good

JamesSmartCell:
It's possible you need to do the login first. The exact sequence I did was also to test metamask, I got an eth from the faucet, logging in in the process. Then it should work with that account

hboon:
@JamesSmartCell ah, thanks. It works in iOS already. The trick like you said is to create an account and login first to reproduce it properly.

PR #862, opened 2019-09-14
Release 2-11-0

JamesSmartCell:


James-Sangalli:
@JamesSmartCell please use tags in future, you can push a version number change to master and tag it if you desire (just don't put anything else in it)

PR #865, opened 2019-09-17
Refactor Action Button bar 

JamesSmartCell:
This PR refactors the action button bar which had got a bit messy. Created a new drop-in class that's a single drop-in and setup for any screen using TokenScript actions.


JamesSmartCell:
The button bar is now a generic drop-in for all contract types that can grow to any size. @colourful-land please push through I will release after testing.

James-Sangalli:
@JamesSmartCell would this solve #870?

PR #869, opened 2019-09-22
auto adds the discovery token

James-Sangalli:
WIP because while this is called it doesn't actual add the token like the code would suggest @JamesSmartCell please advise


James-Sangalli:
@JamesSmartCell ready to go

James-Sangalli:
related? https://github.com/AlphaWallet/alpha-wallet-android/issues/870

PR #875, opened 2019-09-24
Fix token detection when reading constructor bytes that have a proxy contract inside it

JamesSmartCell:
Was failing previously due to Token contract with unassigned proxy.

Now uses the following heuristic:

Call getBalance.
- result is a single integer: Assume ERC20 (If ERC721 then OpenSea call will override this).
- result is an array declaration with 1 or more entries: Assume ERC875, and set off thread to determine if ERC875 contract is legacy.
- result is neither of these: Contract doesn't contain any balance query; unknown contract attempt to query constructor.

### DIFF
--- a/app/src/main/java/io/stormbird/wallet/interact/FetchTransactionsInteract.java
+++ b/app/src/main/java/io/stormbird/wallet/interact/FetchTransactionsInteract.java
@@ -57,7 +57,6 @@ public class FetchTransactionsInteract {
     {
         switch (type)
         {
-            default:
             case ERC20:
             case ERC721:
                 return Single.fromCallable(() -> type);
@@ -66,6 +65,10 @@ public class FetchTransactionsInteract {
                 transactionRepository.queryInterfaceSpec(tokenInfo.address, tokenInfo)
                         .subscribe(actualType -> TokensService.setInterfaceSpec(tokenInfo.chainId, tokenInfo.address, actualType)).isDisposed();
                 return Single.fromCallable(() -> type);
+            default:
+                //Token wasn't any of the easily determinable ones, use constructor to analyse
+                return tokenRepository.resolveProxyAddress(tokenInfo) //resolve proxy address to find base constructor and analyse
+                        .flatMap(address -> transactionRepository.queryInterfaceSpec(address, tokenInfo));
         }
     }
 
--- a/app/src/main/java/io/stormbird/wallet/interact/FetchTransactionsInteract.java
+++ b/app/src/main/java/io/stormbird/wallet/interact/FetchTransactionsInteract.java
@@ -63,6 +63,7 @@ public class FetchTransactionsInteract {
             case ERC875:
                 //requires additional handling to determine if it's Legacy type, but safe to return ERC875 for now:
                 transactionRepository.queryInterfaceSpec(tokenInfo.address, tokenInfo)
+                        .subscribeOn(Schedulers.io())
                         .subscribe(actualType -> TokensService.setInterfaceSpec(tokenInfo.chainId, tokenInfo.address, actualType)).isDisposed();
                 return Single.fromCallable(() -> type);
             default:
--- a/app/src/main/java/io/stormbird/wallet/interact/FetchTransactionsInteract.java
+++ b/app/src/main/java/io/stormbird/wallet/interact/FetchTransactionsInteract.java
@@ -1,5 +1,6 @@
 package io.stormbird.wallet.interact;
 
+import io.reactivex.SingleSource;
 import io.stormbird.wallet.entity.*;
 import io.stormbird.wallet.repository.TokenRepositoryType;
 import io.stormbird.wallet.repository.TransactionRepositoryType;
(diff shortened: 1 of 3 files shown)


James-Sangalli:
@JamesSmartCell an unassigned proxy? could you please elaborate further.

JamesSmartCell:
The contract has the 'implementation' function which is reserved for adjusting proxy contract throws. This function returns the zero address (specifically), so it was confusing the function that normally detects the contract at the throw address. Changing to a heuristic gives better and quicker results.

What will the 'implementation' function do in this contract?

James-Sangalli:
@JamesSmartCell Sorry what contract?

Also, I am not sure what the null address has to do with this?

JamesSmartCell:
The new ALP discovery one. Anyway, this code works a lot better than before in the vast majority of cases. It uses the older code now mainly just to determine if it's a legacy type contract.

JamesSmartCell:
Wallets have to detect the contract type when they get a token added. I don't understand why you don't understand the need for this ...

hboon:
So when the user adds a custom token, there is heuristics applied to determine the token type. One of it is to query the constructor (maybe to search for specific function signature hashes?)

And this PR fixes an issue with the heuristics when applied to a contract which is meant to be a proxy, and in this case is triggered by the discover token.

@JamesSmartCell does that sum it up?

James-Sangalli:
> Wallets have to detect the contract type when they get a token added. I don't understand why you don't understand the need for this ...

That means something but you are talking about some kind of proxy contract that throws and none of that means anything to me. Please explain it. 

Ideally a PR like this would clearly explain what the fix is and how it was done, it is not at all clear in this PR.

JamesSmartCell:
Did you have the 'implemetation' function in your ALP Discovery contract? My code was calling this and it returned a positive, although it returned the zero address. 'implementation' is usually reserved for proxy contracts. The most likely reason is that you specified this so that in future if you needed to change the contract you can set up a new contract that this one forwards function calls to.

James-Sangalli:
> Did you have the 'implemetation' function in your ALP Discovery contract? My code was calling this and it returned a positive, although it returned the zero address. 'implementation' is usually reserved for proxy contracts. The most likely reason is that you specified this so that in future if you needed to change the contract you can set up a new contract that this one forwards function calls to.

What is an implementation function?

The code is here: https://github.com/AlphaWallet/contracts/blob/master/AlphaWallet-Discover/AlphaWalletDiscover.sol

There is a proxy contract in there but it is only called inside a function to execute multiple transfers in one via the transferFrom function signature. It is not implemented however as this would make the contract be like this: 

`AlphaWalletDiscover  is Proxy` but this didn't happen

JamesSmartCell:
Ok I see; that proxy must have been throwing the constructor detection code. I'd been planning to add this optimisation for a while since for 99% of contracts it'll be 20, 721 or 875. It's better to depend on a basic Infura call than fetching from etherscan. Either way; the optimisation won't trip up on this type of contract.

JamesSmartCell:
See top comment for explanation of PR. 

Due to the definition of the ALP contract it was confusing the constructor analysis code. I refactored to fix this code and provide improved token type resolve speed.

James-Sangalli:
Ok now I understand, is there a more robust way to detect however?

JamesSmartCell:
Initially looking at the constructor worked very well but it failed with proxied contracts since it was analysing the constructor of the proxy not the token itself (eg DAI).

To keep the pure constructor analysis method you'd need to check first the base address constructor, see if it's a valid token then check the proxied token.

I think it's a lot more robust to use a heuristic, considering we only show ERC20, 721 and 875:

- first get balance. If it's an integer then we have a token, which is either ERC20 or ERC721. Since we call opensea we can correct the type when it gets pulled in from opensea. If it's an array balance we know it's erc875. It's not urgent to know if it's erc875 legacy (and increasingly less relevant), so that is determined in the background via the constructor analysis.

Currently there's (stupidly) no single method to determine token type other than trying the interface method. Does that consume gas? Problem is we don't implement it - or do we now? This would be the best system.

James-Sangalli:
> Initially looking at the constructor worked very well but it failed with proxied contracts since it was analysing the constructor of the proxy not the token itself (eg DAI).
> 
> To keep the pure constructor analysis method you'd need to check first the base address constructor, see if it's a valid token then check the proxied token.
> 
> I think it's a lot more robust to use a heuristic, considering we only show ERC20, 721 and 875:
> 
> * first get balance. If it's an integer then we have a token, which is either ERC20 or ERC721. Since we call opensea we can correct the type when it gets pulled in from opensea. If it's an array balance we know it's erc875. It's not urgent to know if it's erc875 legacy (and increasingly less relevant), so that is determined in the background via the constructor analysis.
> 
> Currently there's (stupidly) no single method to determine token type other than trying the interface method. Does that consume gas? Problem is we don't implement it - or do we now? This would be the best system.

Balance method seems relatively solid, there is eip-165 but few use it so we can't rely on it.

JamesSmartCell:
EIP165; that's the supportsInterface method? It consumes gas AFAIK, which doesn't seem very user friendly.

hboon:
> EIP165; that's the supportsInterface method? It consumes gas AFAIK, which doesn't seem very user friendly.

It doesn't consume gas. It's a function call, not a transaction.

hboon:
This is what we do for iOS. I thought, we did something similar for Android but with additional heuristics?

1. Check if ERC875 by calling isStormBirdContract() and checking for true
2. Check if ERC721 by
    * If it matches these 2 contracts, it is ERC721
        * static let bitizen = "0xb891c4d89c1bf012f0014f56ce523f248a07f714"
        * static let cryptoSaga = "0xabc7e6c01237e8eef355bba2bf925a730b714d5f"
    * Use supportsInterface
        * If 0x9a20483d = CryptoKitty
        * If 0x80ac58cd = Official ERC721
        * If 0x6466353c = old ERC721
3. Else
    ERC20
4. ERC721 tokens reported by OpenSea override the above

JamesSmartCell:
The 'isStormbirdContract()' seems incongrous - what if other people start using it? My logic is that ERC875 is the only token type that returns an array. It's easy to detect an array, the first element will be the array marker (0x20) and the length of the returned value will be greater than one element (ie > string length() 64). If another token type returns a balance array, then it will be treated as ERC875 which is OK for now.

I must have misunderstood the requirement for gas, it seemed strange at the time for a static function.
I'll add the 'supportsInterface' for ERC721 type for transfer then it should be good to go.

hboon:
> The 'isStormbirdContract()' seems incongrous - what if other people start using it? My logic is that ERC875 is the only token type that returns an array.

That seems unlikely unless they were trying to screw things up (or copied our contract, then it's likely an ERC875). If it's the former, it's comparable to returning the wrong hash for supportsInterface?

It might be safer to check for an array for now, but in the future, there might be other token standards doing that without any ill-intentions.

Any way, I don't have a strong stance on either. Just pointing out they both aren't perfect. They both work for now to some degree.

JamesSmartCell:
OK I implemented @hboon 's suggestion to use EIP165. My flow pretty much matches his now except for using balance to distinguish ERC875 instead of isStormbirdContract, and if it's an ERC875 then the constructor analysis will determine ERC875 legacy (really not so important now)

PR #881, opened 2019-09-25
Fix: changes not loaded from onConfirm callback & add more input types 

JamesSmartCell:
- Fix for different types of input.
- ensure 'onConfirm' is called before evaluating input types.


JamesSmartCell:
It's ok if you think it makes the intention more clear. Previously I hadn't called 'onConfirm' if the TokenScript function was a transaction call request. But, a developer may expect or need onConfirm to be called before evaluating inputs as you discovered - there could be other conditions this is needed.

PR #911, opened 2019-10-03
Implement EIP155 signing

JamesSmartCell:
Implement EIP155 signing to ensure sigs from one chain can't be used in another.
TODO: Fully implement EIP155 when web3j 4.5.0 for Android is published (chainId is upgraded from 1 byte to 8 bytes in order to implement chainId's greater than 255).

@hboon Recall our conversation about EIP155 usage in Android - in summary previously chainId wasn't used in the signature. The current implementation in web3j 4.2.0 treats the 'V' in the signature as a single byte as per the original Ethereum implementation. When it upgrades to 4.5.0 we can use the full 8 byte 'long' implementation but until then this implementation uses the intermediate single byte which at least protects all the original chains and testnets.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/repository/TransactionRepository.java
+++ b/app/src/main/java/com/alphawallet/app/repository/TransactionRepository.java
@@ -34,6 +34,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import static com.alphawallet.app.entity.CryptoFunctions.sigFromByteArray;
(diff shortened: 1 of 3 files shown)


hboon:
@JamesSmartCell related, is paymaster for ticket tokens affected?

JamesSmartCell:
It would be, yes. It should also be fixed in there.

PR #912, opened 2019-10-03
Fix TokenScript Issues related to Compound functions

### LINKED ISSUE #909: Webview for TokenScript's does not update to changes in the JS code
Relevant #901 & #906 (probably the cause)

For example, the TokenScript's JS code might check that the user has enabled by checking the this.props.allowance > 0, if it is then the screen will show the input box and allow the user to do a withdraw/deposit.

The client simply only loads the first view and doesn't update with this new information that comes from a smart contract 

See example file: https://github.com/AlphaWallet/TokenScript/blob/master/examples/erc20/DeFi/Compound/enable.en.shtml#L10

JamesSmartCell:
- Ensure TokenScript Ethereum functions update correctly.
- Fix issue with duplicate JavaScript call.
- Add window.onClose listener.

Fixes #909

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokenscript/TokenScriptRenderCallback.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokenscript/TokenScriptRenderCallback.java
@@ -2,5 +2,5 @@ package com.alphawallet.app.entity.tokenscript;
 
 public interface TokenScriptRenderCallback
 {
-    void callToJSComplete(String function);
+    void callToJSComplete(String function, String result);
 }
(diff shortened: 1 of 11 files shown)


James-Sangalli:
@JamesSmartCell must the window.onClose callback be defined explicitly? It is part of the webview JS

JamesSmartCell:
Yes, it's the only way to intercept the window.onClose on Android. You can't get it from the JS execution callback. One alternative which would be neater would be to define some standard pass/fail codes which return the relevant code to indicate pass/fail from the 'onConfirm'.

Even if it's just returning a string of "1" to indicate pass and "0" for fail it we could pick those up and respond accordingly.

JamesSmartCell:
OK last part should be working now - that is, if a window.onClose happens when a button is pressed, don't perform any subsequent action, just close the action card.

Issue #903, opened 2019-10-02
NPE on CoinmarketCapTicketService apply()


James-Sangalli:
java.lang.NullPointerException: Attempt to invoke interface method 'void io.reactivex.disposables.Disposable.dispose()' on a null object reference
        at io.reactivex.internal.observers.BasicFuseableObserver.fail(BasicFuseableObserver.java:109)
        at io.reactivex.internal.operators.observable.ObservableMap$MapObserver.onNext(ObservableMap.java:59)
        at com.alphawallet.app.service.CoinmarketcapTickerService$ApiErrorOperator$1.onNext(CoinmarketcapTickerService.java:85)
        at com.alphawallet.app.service.CoinmarketcapTickerService$ApiErrorOperator$1.onNext(CoinmarketcapTickerService.java:77)
        at retrofit2.adapter.rxjava2.CallExecuteObservable.subscribeActual(CallExecuteObservable.java:47)
        at io.reactivex.Observable.subscribe(Observable.java:12267)
        at io.reactivex.internal.operators.observable.ObservableLift.subscribeActual(ObservableLift.java:57)
        at io.reactivex.Observable.subscribe(Observable.java:12267)
        at io.reactivex.internal.operators.observable.ObservableMap.subscribeActual(ObservableMap.java:32)
        at io.reactivex.Observable.subscribe(Observable.java:12267)
        at io.reactivex.internal.operators.observable.ObservableSubscribeOn$SubscribeTask.run(ObservableSubscribeOn.java:96)
        at io.reactivex.Scheduler$DisposeTask.run(Scheduler.java:578)
        at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
        at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:301)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)

PR #924, opened 2019-10-07
Use Amberdata API

JamesSmartCell:
Use Amberdata API as primary data source. This allows us to add tickers for ERC20 tokens that have a presence in the Amberdata market price database.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,8 +38,8 @@ android {
         applicationId "io.stormbird.wallet"
         minSdkVersion 23
         targetSdkVersion 29
-        versionCode 76
-        versionName "2.12.1"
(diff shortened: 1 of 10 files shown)


JamesSmartCell:
I sent you a PM about where to get the private API key.

JamesSmartCell:
Still it demonstrates the fallback pricing which is obtained from the DAI oracle works fine, in case both the APIs are down.

James-Sangalli:
@JamesSmartCell closes https://github.com/AlphaWallet/alpha-wallet-android/issues/903?

JamesSmartCell:
good catch.
Fixes #903 .

PR #940, opened 2019-10-14
Fix CORS web3 error

JamesSmartCell:
Fix for web3 provider not found issue as seen when trying to use https://1inch.exchange (reported on the Alphawallet telegram group). This was caused by the Android webview not allowing the Cross Origin call from within the site. To fix I added cross origin allowed header.


James-Sangalli:
@JamesSmartCell I thought it was up to the server not the client to allow CORS calls? https://stackoverflow.com/questions/36958999/cors-is-it-a-client-side-thing-a-server-side-thing-or-a-transport-level-thin

JamesSmartCell:
There's calls originating from within the web page (JavaScript calls to external servers). The default webview blocks them. Without those headers these calls fail due to security policies within Android webview. iOS doesn't have those restrictions.

hboon:
>@JamesSmartCell I thought it was up to the server not the client to allow CORS calls?

@James-Sangalli server states its preference, user agent (browser) enforces it. Think about when you use curl.

PR #952, opened 2019-10-27
Fix opensea attributes not showing up

### LINKED ISSUE #948: Should get all attributes from opensea into token instance
Android screenshot, should have the attributes in the last screenshot:
Screenshot 2019-10-22 21 15 15

iOS: has the opensea details in full but doesn't include the attributes from the tsml (bug reported here: https://github.com/AlphaWallet/alpha-wallet-ios/issues/1504)

Screenshot 2019-10-22 21 15 25

James-Sangalli:
fixes #948

### DIFF
--- a/app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
+++ b/app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
@@ -1207,7 +1207,7 @@ public class AssetDefinitionService implements ParseResult, AttributeInterface
         }
     }
 
-    private void addOpenSeaAttributes(StringBuilder attrs, ERC721Token erc721Token, BigInteger tokenId)
+    private void addOpenSeaAttributes(StringBuilder attrs, Token erc721Token, BigInteger tokenId)
     {
         Asset tokenAsset = erc721Token.getAssetForToken(tokenId.toString());
         if(tokenAsset == null) return;
@@ -1240,7 +1240,7 @@ public class AssetDefinitionService implements ParseResult, AttributeInterface
 
         if(token instanceof ERC721Token)
         {
-            addOpenSeaAttributes(attrs, (ERC721Token) token, tokenId);
+            addOpenSeaAttributes(attrs, token, tokenId);
         }
 
         if (token.isEthereum())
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
```diff
@@ -1203,6 +1238,11 @@ public StringBuilder getTokenAttrs(Token token, BigInteger tokenId, int count)
         TokenScriptResult.addPair(attrs, "tokenId", tokenId);
         TokenScriptResult.addPair(attrs, "ownerAddress", token.getWallet());
```
JamesSmartCell: This should be done with OO abstraction, also don't cast to ERC721 token just pass the token directly.

You could do something like this at the base level:

```
if (token instanceof ERC721Token) 
{
   addOpenSeaAttributes(attrs, token, tokenId);
}
```
Add the function getAssetForToken into class Token, return null, then specify it's an @Override in the ERC721Token definition.

app/src/main/java/com/alphawallet/app/service/OpenseaService.java
```diff
@@ -78,7 +78,7 @@ public OpenseaService(Context ctx) {
                     foundTokens.put(asset.getAssetContract().getAddress(), token);
                 }
 
-                ((ERC721Token) token).tokenBalance.add(asset);
+                ((ERC721Token) token).addAssetToTokenBalanceAssets(asset);
```
JamesSmartCell: This should be abstracted as above.

app/src/main/java/com/alphawallet/app/ui/widget/adapter/NonFungibleTokenAdapter.java
```diff
@@ -135,7 +135,7 @@ protected void setERC721Tokens(Token token, String ticketId)
         int weight = 1; //use the same order we receive from OpenSea
 
         // populate the ERC721 items
-        for (Asset asset : ((ERC721Token)token).tokenBalance)
+        for (Asset asset : ((ERC721Token)token).getTokenAssets())
```
JamesSmartCell: Abstract

app/src/main/java/com/alphawallet/app/viewmodel/TransferTicketDetailViewModel.java
```diff
@@ -252,7 +252,7 @@ public void openConfirm(Context ctx, String to, Token token, String tokenId, Str
     {
         //first find the asset within the token
         Asset asset = null;
-        for (Asset a : ((ERC721Token) token).tokenBalance)
+        for (Asset a : ((ERC721Token) token).getTokenAssets())
```
JamesSmartCell: Abstract

James-Sangalli:
@JamesSmartCell see latest commit

PR #953, opened 2019-10-28
add a default bitmask value for TokenScript handling

James-Sangalli:


James-Sangalli:
@JamesSmartCell yes, since the variable is fixed I think it is easier to read in this format.

Issue #718, opened 2019-06-14
With a new install, ALL is empty until we tap to CURRENCY and back


hboon:
i.e. the Ethereum card doesn't show up.

~~Doesn't happen if I create a new wallet.~~ [updated to remove confusion]

Xiaomi. 7.1.2 N2G47H.

colourful-land:
> Doesn't happen if I create a new wallet.

Since you always create a new wallet with a new install, I guess you mean "cease to happen when I created a𝑛𝑜𝑡ℎ𝑒𝑟 new wallet"?

hboon:
No, the original title meant to convey that this bug is triggered for both (A) new installs and (B) creating another new wallet at any point after that. But it turns out that I can't reproduce (B) anymore. So I have updated the title to say that it is only for new install.

JamesSmartCell:
Does it work now? IE after the update? I can't reproduce it.

hboon:
I still observe it with the version on the play store as of 20190701.

<img src="https://user-images.githubusercontent.com/56189/60441598-e597fe80-9c49-11e9-84bd-54b3d295f85e.jpg" width=200>

I noticed that after a while (I took a photo of the screenshot, cropped and shared it with my laptop), the device was locked and when unlocked, the 0 ETH ether card appears. Maybe it takes a while to appear?

ronakdave195:
@JamesSmartCell May I request you to provide test steps and more description here so that we can try to reproduce this at our end to suggest any fixes.

JamesSmartCell:
This is fixed, I believe @hboon said he doesn't see it anymore.

James-Sangalli:
Tested with later builds on emu, problem is gone

PR #959, opened 2019-11-05
Fix: new wallets not showing cards due to filters

JamesSmartCell:
Issue with new wallet not showing any cards due to filters init.


James-Sangalli:
@JamesSmartCell fixes https://github.com/AlphaWallet/alpha-wallet-android/issues/718

Issue #941, opened 2019-10-14
Add ability to remove and restore tokens


JamesSmartCell:
Extend the code that currently allows sliding out of the backup reminders to tokens. If possible do something like in the Android notifications - expose a garbage icon on the slide and only remove if the user clicks the garbage or hide icon. If tokens are hidden set the database:

ChangeTokenEnableInteract.setEnable - feel free to refactor this to accept a Boolean or whatever.

Add a 'Token Management' entry in the settings page which allows users to bring their tokens back.

Don't worry about UI appearance for now - just blend with existing. Once Tomek the UX overhaul is done it should be simple to migrate to look like that.

PR #963, opened 2019-11-05
WIP : Swipe Tokens to Remove + Token Management Setting

justindg:
Closes https://github.com/AlphaWallet/alpha-wallet-android/issues/941

There are still some issues with this feature as mentioned to @JamesSmartCell.

- Some tokens still show up despite being enabled=false in Realm
- Janky/buggy refresh when removing items
- DAI Stablecoin v1 weirdly reverting to enabled=true even after disabling/swiping it


justindg:
Too outdated to continue; creating another PR instead

PR #966, opened 2019-11-08
handles spawnables via paymaster

James-Sangalli:
Note: call on port 80 has been removed a while ago...


James-Sangalli:
@JamesSmartCell let's address those in a later sprint

PR #975, opened 2019-11-15
Show import progress for seed import

JamesSmartCell:
Add in progress spinner for import seed.


hboon:
@JamesSmartCell does it take that long to import a seed?

JamesSmartCell:
I think it only showed up because I was debugging, and Android debugging seems to have moved backwards over the past few months - it's now much slower than it was last year, maybe changes to Android Studio or ADB target connection manager. 

Even so someone's Android phone could be lagging or running slow and they'd get a pause where nothing appears to be happening. There should be a progress spinner.

PR #977, opened 2019-11-21
add ERC721 ticket token type and smart contract function calls

### LINKED ISSUE #981: get balances from special mapping for ERC721 ticket tokens

James-Sangalli:
Closes #981

@JamesSmartCell need your careful attention on this PR.

This PR:
- Adds ERC721 ticket token type to enum
- Retrieves whether it is this type by calling ERC165 function for supported interface "getBalances(address)"
- Adds itself to the ticket.java class and handles unique cases
- Gets it's own balance without opensea by calling getBalances(address) if the type is ERC721_TICKET
- Renames a few related things 
- Moves token classes into its own package (@JamesSmartCell this is the major cause for some many files changing but the changes are only the package name so that shouldn't make this PR hard to review)

Still need to:
- Use this enum to create custom magic links with token id
- Use this enum to create a unique redeem QR code which has the token id


James-Sangalli:
@JamesSmartCell please put your ticket type fixes here

PR #1000, opened 2019-11-27
Abstract grouping heuristic

JamesSmartCell:
Abstract the grouping heuristic to the respective token class. This gives fine control over how tokens are displayed in the full asset card view (```view``` in TokenScript).

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokens/ERC721Ticket.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokens/ERC721Ticket.java
@@ -11,6 +11,7 @@ import android.widget.TextView;
 import com.alphawallet.app.R;
 import com.alphawallet.app.entity.ContractType;
 import com.alphawallet.app.entity.ERC875ContractTransaction;
+import com.alphawallet.app.entity.TicketRangeElement;
 import com.alphawallet.app.entity.Transaction;
 import com.alphawallet.app.entity.TransactionOperation;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/alphawallet/app/entity/tokens/ERC721Ticket.java
```diff
@@ -411,7 +412,7 @@ public boolean hasArrayBalance()
     }
 
     @Override
-    public List<BigInteger> getArrayBalance() { return balanceArray; }
```
James-Sangalli: @JamesSmartCell under what circumstance would you want the balance with zeros in the array for 721 tickets?

James-Sangalli:
@JamesSmartCell Added some fixes to this PR because it crashed on transfer and I edited it so that a non zero balances in the array are returned to prevent incorrect balance count on token card

app/src/main/java/com/alphawallet/app/entity/tokens/ERC721Ticket.java
JamesSmartCell: This is for erc875. Most likely it's an artifact from the ticket class.

JamesSmartCell: But it should be in a different PR as it's unrelated to this one

James-Sangalli:
@JamesSmartCell remember to reference the issue: https://github.com/AlphaWallet/alpha-wallet-android/issues/987

PR #1001, opened 2019-11-28
Fix for incorrect type overrides

JamesSmartCell:
Ensure determined token type is only overridden by TokenScript definition if the type is incompatible. IE if TS definition says it's ERC721 and currently the determined type is ERC20 then override.

TODO: remove redundant ```"erc721Ticket"``` type in the ```checkCorrectInterface``` method, if it is not going to be used.


app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
```diff
@@ -1058,18 +1058,21 @@ private void checkCorrectInterface(Token token, String contractInterface)
...
                 break;
             case "erc20":
                 cType = ContractType.ERC20;
                 break;
             case "erc721":
                 cType = ContractType.ERC721;
+                if (token.isERC721() || token.isERC721Ticket()) return;
                 break;
             case "erc721Ticket":
                 cType = ContractType.ERC721_TICKET;
```
James-Sangalli: @JamesSmartCell the tokenscript will never be marked as an erc721Ticket

JamesSmartCell: See 'TODO' above. I don't think that value will be used.

PR #1017, opened 2019-12-04
check if spawnables are available

James-Sangalli:
Closes https://github.com/AlphaWallet/alpha-wallet-android/issues/1005

This PR:
- Checks if token is not owned (if created by issuer), if not then it is unavailable (for erc721 tickets)
- If not created by the issuer, checks that the signer actually has the balance (for erc721 tickets)
- Refactors some nasty code


James-Sangalli:
@JamesSmartCell ready to merge

JamesSmartCell:
There's a hardcoded addr marked with a TODO - is this supposed to be in the PR?

PR #1027, opened 2019-12-09
Fix ERC721 transfer display

JamesSmartCell:
Fixes ERC721 token display to ensure consistency between opensea/tokenscript display for standard ERC721 tokens (eg cryptokitty).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -70,8 +70,8 @@ android {
     productFlavors {
 	awallet {
 		dimension "targetting"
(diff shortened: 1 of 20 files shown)


James-Sangalli:
@JamesSmartCell can you be more specific? Not clear what this PR actually does

JamesSmartCell:
In the patch to simplify NFT display I had broken ERC721 opensea display. This fixes it.

PR #1028, opened 2019-12-09
Fix token update janking

JamesSmartCell:
This PR stores erc20 tickers so the wallet is aware the token will be displaying a ticker at startup. This stops the 'janking' caused by erc20 tokens having an additional field appearing for ticker.
It also updates the name of the token from blockscout so that the tokens don't get moved around/renamed after the blockscout refresh.

Both these issues are cosmetic only but caused the token display to jump around after refresh which doesn't look good.

'Janking' is the technical term coined by Justin to describe the issue.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -70,8 +70,8 @@ android {
     productFlavors {
 	awallet {
 		dimension "targetting"
-        versionCode 83
-        versionName "2.20.0"
+        versionCode 82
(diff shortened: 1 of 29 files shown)


JamesSmartCell:
I always get into difficulty when I git rebase - maybe I'm not doing it right but it seems to take ages going through manually editing each commit in an endless loop.

PR #1031, opened 2019-12-11
Tidy up the token checking

JamesSmartCell:
Small PR to clean up some untidy code.

- Tidy up code
- Remove redundant methods
- Only check unknown tokens at startup

### DIFF
--- a/app/src/main/java/com/alphawallet/app/viewmodel/WalletViewModel.java
+++ b/app/src/main/java/com/alphawallet/app/viewmodel/WalletViewModel.java
@@ -286,9 +286,6 @@ public class WalletViewModel extends BaseViewModel
                     .subscribeOn(Schedulers.io())
                     .observeOn(AndroidSchedulers.mainThread())
                     .subscribe(this::storedTokens, this::onError).isDisposed();
-
-            //TODO: Once we start receiving ERC20 tickers check backup requirement if ERC20 received
-            //checkBackup();
         }
     }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/alphawallet/app/viewmodel/WalletViewModel.java
```diff
@@ -315,15 +312,21 @@ private void onFetchTokensCompletable()
...
             balanceTimerDisposable = Observable.interval(0, BALANCE_CHECK_INTERVAL_MILLIS, TimeUnit.MILLISECONDS)
                     .doOnNext(l -> checkBalances()).subscribe();
         }
     }
 
+    private void AddUnresolvedContracts()
+    {
+        Observable.fromArray(ethereumNetworkRepository.getAllKnownContracts(tokensService.getNetworkFilters()).toArray(new ContractResult[0]))
+                .filter(result -> tokensService.getToken(result.chainId, result.name) == null)
+                .forEach(r -> unknownAddresses.add(r)).isDisposed();
```
hboon: Is the `isDisposed()` at the end necessary?

app/src/main/java/com/alphawallet/app/viewmodel/WalletViewModel.java
```diff
@@ -315,15 +312,21 @@ private void onFetchTokensCompletable()
...
-        checkUnknownAddresses();
+        AddUnresolvedContracts();
         if (balanceTimerDisposable == null || balanceTimerDisposable.isDisposed())
         {
             balanceTimerDisposable = Observable.interval(0, BALANCE_CHECK_INTERVAL_MILLIS, TimeUnit.MILLISECONDS)
                     .doOnNext(l -> checkBalances()).subscribe();
         }
     }
 
+    private void AddUnresolvedContracts()
```
hboon: Matter of style. Start `AddUnresolvedContracts()` with a lowercase instead?

app/src/main/java/com/alphawallet/app/viewmodel/WalletViewModel.java
JamesSmartCell: Nicely spotted. It's not necessary, but it's a harmless way to stop Java lint from complaining that the return is not used - which we never need to use in these cases.

app/src/main/java/com/alphawallet/app/viewmodel/WalletViewModel.java
JamesSmartCell: Yes probably.

PR #1035, opened 2019-12-12
Tokenscript Rendering

JamesSmartCell:
Finally fixes the issue with rendering webviews within a listview.

Short version:

you need to determine the final height of the webview before you start drawing it in the list, because the background renders at a different time to the foreground.

Do this by hiding a webview and populating a ticket, measure the webview height then start populating the listview.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
@@ -4,7 +4,9 @@ import android.content.Context;
 import android.os.Parcel;
 import android.os.Parcelable;
 import android.text.TextUtils;
+import android.util.Base64;
 import android.view.View;
+import android.widget.ProgressBar;
 
 import com.alphawallet.app.R;
 import com.alphawallet.app.entity.ContractType;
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
```diff
@@ -108,10 +118,11 @@ protected Token(Parcel in) {
         lastTxCheck = in.readLong();
         lastTxUpdate = in.readLong();
         lastTxTime = in.readLong();
-        int readTS = in.readByte();
-        int readDTS = in.readByte();
-        hasTokenScript = readTS == 1;
-        hasDebugTokenscript = readDTS == 1;
+        hasTokenScript = in.readByte() == 1;
+        hasDebugTokenscript = in.readByte() == 1;
+        nonIconifiedWebviewHeight = in.readInt();
```
hboon: Is this handled for a given token?

1. Has existing TokenScript 
2. Computed height
3. Change TokenScript so that the iconified view has a different height
4. View is rendered with the new height

app/src/main/java/com/alphawallet/app/web3/JsInjectorClient.java
```diff
@@ -203,8 +203,9 @@ private String loadInitJs(Context context) {
 
     String injectStyle(String view, String style)
     {
-        style = "<style type=\"text/css\">\n" + style + "</style>\n";
-        return style + view;
+        String injectHeader = "<head><meta name=\"viewport\" content=\"width=device-width, user-scalable=yes\" /></head>";
```
hboon: I thought we already have something like this somewhere else in Android in order to get the width to fit (`device-width`). But anyway, on iOS, we use:

```
<meta name="viewport" content="width=device-width, initial-scale=1,  maximum-scale=1, shrink-to-fit=no">
```

The main difference is, it's effectively `user-scalable=false`. Users shouldn't be able to treat it like a browser.

app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
JamesSmartCell: This is so when we pass the token between processes anything setup for that token is passed to that process.

However I think you're asking - do I cache the height - I thought about that - needs to be attached to the hash of the script.

app/src/main/java/com/alphawallet/app/web3/JsInjectorClient.java
JamesSmartCell: ah ok. I'll try that

app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
hboon: > However I think you're asking - do I cache the height - I thought about that - needs to be attached to the hash of the script.

@JamesSmartCell yes :)

iOS caches the height based on the script hash and will refresh with the updated height, but unfortunately, it wouldn't work correctly if the view changes height depending on its contents. But I'll just leave it (iOS) for now, because performance is quite bad on it with heavier views without caching.

PR #1039, opened 2019-12-13
New redeem QR format

JamesSmartCell:
Update the QR format for ERC721 tickets.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/SignaturePair.java
+++ b/app/src/main/java/com/alphawallet/app/entity/SignaturePair.java
@@ -2,6 +2,10 @@ package com.alphawallet.app.entity;
 
 import org.web3j.utils.Numeric;
 
+import java.io.ByteArrayInputStream;
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/alphawallet/app/entity/SignaturePair.java
```diff
@@ -57,13 +61,25 @@ public String formQRMessage()
...
     public static String generateSelection721Tickets(List<BigInteger> tokenIds) {
+        String prefix = "0"; //future proof the format - this is one digit to refer to token count or protocol format
         StringBuilder stringBuilder = new StringBuilder();
         for(BigInteger token: tokenIds) {
-            stringBuilder.append(Numeric.toHexStringNoPrefix(token));
+            stringBuilder.append(token.toString(10));
             stringBuilder.append(",");
         }
-        return stringBuilder.substring(0, stringBuilder.length() - 1);
+        String tokenIdsString = stringBuilder.substring(0, stringBuilder.length() - 1);
```
James-Sangalli: @JamesSmartCell Stringbuilder.toString?

JamesSmartCell: I don't understand what this comment means.

James-Sangalli: Why use substring when you can just use the string directly? I.e. you are using substring for the full length so there is no need to substring

JamesSmartCell: I did minimum of changes to the code you added ... I guessed that you did the substring because eventually it could handle multiple tokens.

PR #1040, opened 2019-12-14
Check ticket redeem

JamesSmartCell:
Now does a periodic check to see if an ERC721 ticket has been redeemed.


James-Sangalli:
@JamesSmartCell this is handled in the tokenscript, why the need for new handling code specifically for redeem?

JamesSmartCell:
So the notification completes when you have the 'redeem' screen open. It will also update in tokenscript once the new transaction is seen.

PR #1043, opened 2019-12-15
More generification options

JamesSmartCell:
More generification options


app/src/main/res/layout/layout_function_buttons.xml
```diff
@@ -13,7 +13,7 @@
         android:layout_width="match_parent"
         android:layout_height="?actionBarSize"
         android:layout_gravity="center"
-        android:background="@color/nasty_green"
```
James-Sangalli: @JamesSmartCell this is relevant to keeping the code base more generic?

JamesSmartCell: yes that's right. By having a consistent button colour it's easy to universally change the button colour instead of setting 'nasty_green' to whatever colour you want the buttons to be, which isn't even used everywhere.

PR #1047, opened 2019-12-16
Added SpiderDEX

TamirTian:


TamirTian:
I rechecked it and i think it's ok.

Let me know if there's anything else I can do.

JamesSmartCell:
It's fine. The error is at our end, I added a nice greeting for first commit but the greeting fails 😂. I'll remove it and merge.

PR #1057, opened 2020-01-07
Recover incorrect cipher

JamesSmartCell:
Fix for Issue caused by incorrect key classification. The key was classified as keystore when it should have been a legacy keystore. This patch pre-parses the keys to identify the key type at startup.

The issue was reported on Twitter.


JamesSmartCell:
Yes, we have to handle legacy keys from the original wallet which are stored using a lower strength cipher - as identified during our crypto review.

PR #1058, opened 2020-01-07
Fix Tokenscript Persistence

JamesSmartCell:
- Initiate asset map reload if OS has scavenged it. This happens if user pages out, performs a memory intensive task like looking at a map or playing a game then paging back in. Previously, if OS scavenged the asset map the user would now be looking at a blank token instead of the tokenscript view or iconified.

- Refactor mapping to use the more efficient SparseArray (as recommended practice).


James-Sangalli:
@JamesSmartCell will this cache values which may have updated in the meantime?

JamesSmartCell:
@James-Sangalli that's not the purpose of this PR, this is only to refresh the file map so any TS enabled token that's being used will be displayed correctly; RE cached values, currently any cached values will be updated if any transactions have been written to the contract. This is in lieu of an event listener and parser.

PR #1059, opened 2020-01-08
Change node connection for MainNet and Rinkeby.

JamesSmartCell:
Switch mainnet and rinkeby connections:

- Use Amberdata for MainNet and Rinkeby node connections.
- Use Amberdata for token discovery source (ie one call to find all market listed ERC20 tokens on mainnet).

Note this is an intermediate step toward the structured node endpoint.


JamesSmartCell:
Redundant, ticker using Amberdata was done in a standalone PR. Mainnet node switched to AmberData in previous PR.

PR #1065, opened 2020-01-12
prevent java pid hprof files from being commited

James-Sangalli:
Sometimes my computer creates these massive files and tries to add them


colourful-land:
It's included in .gitignore just a few minutes ago. Sorry I didn't notice this PR.

PR #1072, opened 2020-01-14
Fix for when tokenscript function returns null

JamesSmartCell:
If Tokenscript function has no return then integer output should be zero.


James-Sangalli:
https://github.com/AlphaWallet/shankai-UEFA-Android/issues/41

PR #1088, opened 2020-01-21
Allow uploading of files in the dapp browser 

JamesSmartCell:
Before this PR, it was not possible to upload files when visiting websites in the dapp browser. This PR allows users to upload their files.


James-Sangalli:
@JamesSmartCell please elaborate as this is very vague

PR #1089, opened 2020-01-22
Fix Tokenscript Result Fetch

### LINKED ISSUE #1087: TokenScripts don't update after conditional on props
Example:

Compound requires you enable the dai contract before using supply or withdraw, our TokenScript checks if you have enabled and if so, allows you to supply/withdraw and if not displays a message saying you must enable first.

In Android this does not work, when you have updated, the prop check fails to update the screen to allow you to supply/withdraw after checking the allowance prop to see if it is enabled.

JamesSmartCell:
Closes #1087 

Ensure valid result is used even if contract does not have any known transactions.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokenscript/TokenscriptFunction.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokenscript/TokenscriptFunction.java
@@ -583,7 +583,6 @@ public abstract class TokenscriptFunction
             }
             else  //if value is old or there wasn't any previous value
             {
-                //for function query, never need wallet address
                 return fetchResultFromEthereum(walletAddress, useAddress, attr, tokenId, td, attrIf, transactionUpdate)       // Fetch function result from blockchain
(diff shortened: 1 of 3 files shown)


JamesSmartCell:
Stage 1 of this fix (Working, non-optimised pending event handling):

If TokenScript attribute is sourced from a contract then:
 - if contract is the TokenScript token contract, check if there's been any transactions recorded since last result (including transactions not associated with wallet), if not, used cached result, if there's been any new transactions then refresh the value from the contract.
- if contract is not the TokenScript token contract then always take the value from the contract (if network not available or node times out then use last cached value).

Previously, the cached value would override the new value fetched from Ethereum because the external contract hadn't had any transactions recorded so the logic assumed nothing had changed and used the cached value. 

TODO in stage 2: Use event handling to trigger cache invalidation which will force the app to gracefully refresh the values when required. This may not be possible in all circumstances as it requires event interpretation.

Issue #976, opened 2019-11-20
Display suggestions when importing long SEED phrase


tomekalphawallet:
Please use Moonlight app (iOS) as a reference.

![Import_with_Seed_Android](https://user-images.githubusercontent.com/51817359/69209722-6c735680-0bac-11ea-8b08-3ca832ea1cf1.jpg)

tomekalphawallet:
https://github.com/AlphaWallet/alpha-wallet-ios/issues/1568

JamesSmartCell:
Yes we asked some ppl to come up with a UX for this previously - it will make a good project for external developer. I think there's enough info in the above picture for a developer to implement this. The suggestions would come from the valid word list which is part of the project anyway.

JamesSmartCell:
This has been assigned as a task for external team.

jzaki:
![](https://github.trello.services/images/mini-trello-icon.png) [[1400] As a wallet user, I want to import my wallet's SEED phrase more readily, so that I'm not frustrated with typing long seed phrases](https://trello.com/c/b47sz01Q/31-1400-as-a-wallet-user-i-want-to-import-my-wallets-seed-phrase-more-readily-so-that-im-not-frustrated-with-typing-long-seed-phras)

jzaki:
For posterity, the dictionary is just words from [bip39](https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt)

PR #1090, opened 2020-01-22
Display suggestions when importing long SEED phrase

ChintanRathod:
Implement Task #976 

- Add word count badge at the bottom of input field
- Add suggestion list to provide predefined bip39 English words
- Hide suggestion strip of inbuilt keyboard
- Hide suggestion bar at initial till user start typing


ChintanRathod:
@JamesSmartCell Glad to hear that. 

For the first point of multi-lines, previous height of layout was minimum 170 where I changed to 100 as of now to show word count precisely. I was done due to some of devices might face issue to look into word count as keyboard is open and it get covered with it. 
Solution: We can revert the minimum height from 100 to 170 if it is fine to overlap wordcount in smaller device. 

For the second query, it is a good idea to put remaining words count like 1/12, 2/12 ... 11/12 and complete. We can also change color of text to green as well. 

Kindly share your views on this.

JamesSmartCell:
First point, is that for devices with lower res, yeah we struggled to accommodate those too. Perhaps put the word count inside the box at the top corner and have the box scrollable. That way it's not consuming a whole row.

Second point, yes that's a good alternative, doing 1/12 -> 12/12 would work great. Changing colour at 12/12 or making the text bold, anything to indicate completion is good.

colourful-land:
> First point, is that for devices with lower res, yeah we struggled to accommodate those too. Perhaps put the word count inside the box at the top corner and have the box scrollable. That way it's not consuming a whole row.

@ChintanRathod at the end of 2018 we researched our userbase and discovered our users tends to use new devices (0.5% of our users were using api level < 23 where the market average was like 11%). It could be a self-fulfilling prophercy but I feel at this early stage of blockchain it's acceptable to focus on high-end users.

JamesSmartCell:
@ChintanRathod let me know if the above makes sense or not - displaying as a grid may waste too much space and can be done at a different time as it adds extra complexity for small usability gain.

ChintanRathod:
@JamesSmartCell Sorry to revert late due to out of station travel with limited network available. I need to check with Grid view as we might need a design first how we want to show it. There are two possibilities, 
1) with grid and suggested list of words where user has to tap on word without typing. 
2) with grid and suggested list with keyboard search functionality. 
Both has pros and cons with height and space occupancy. 

For now, what I can do is, I can work on that word count 1/12, 2/12 part and spacing of text area.

JamesSmartCell:
> For now, what I can do is, I can work on that word count 1/12, 2/12 part and spacing of text area.

I like this solution. A more consistent interface to the grid solution would be to mimic the display seed style view, where the word pop into the grid inside 'lozenges' when complete. I'll attach a pic, but have a look at the screen you see when you choose to create a new wallet, where the seed phrase words are listed.

Issue #1091, opened 2020-01-24
not asking for authentication when a private key (not a keystore or HD key) is imported


James-Sangalli:
In the codebase, the app won't ask for authentication if a private key was imported and will ask for authentication if it is a keystore or HD key.

This doesn't make sense as the key type is irrelevant, what determines whether authentication is required is determined by whether the user has locked the key on the device or not. 

Furthermore, because of this issue, users are unable to send transactions if they have a keystore but haven't locked it with a PIN.

colourful-land:
![Screenshot_20200126-153229](https://user-images.githubusercontent.com/548435/73130725-fb3f9c00-4051-11ea-89f6-1be5c3447632.jpg)

Tested. Can't reproduce the problem (app does ask for authentication)

PR #1092, opened 2020-01-24
rename to private key

James-Sangalli:
Found this while trying to debug #1091, seems that keys are being authenticated on their type rather than whether they are secured by the enclave or not. This PR simply renames an incorrectly named enum.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/service/KeyService.java
+++ b/app/src/main/java/com/alphawallet/app/service/KeyService.java
@@ -45,7 +45,6 @@ import java.util.List;
 
 import static android.os.VibrationEffect.DEFAULT_AMPLITUDE;
 import static com.alphawallet.app.entity.Operation.*;
-import static com.alphawallet.app.entity.ServiceErrorException.ServiceErrorCode.INVALID_KEY;
 import static com.alphawallet.app.entity.tokenscript.TokenscriptFunction.ZERO_ADDRESS;
 import static com.alphawallet.app.service.KeyService.AuthenticationLevel.STRONGBOX_NO_AUTHENTICATION;
 import static com.alphawallet.app.service.KeyService.AuthenticationLevel.TEE_NO_AUTHENTICATION;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/alphawallet/app/service/KeyService.java
```diff
@@ -254,7 +253,7 @@ public void getAuthenticationForSignature(Wallet wallet, Activity callingActivit
 
         switch (wallet.type)
         {
-            case KEYSTORE_LEGACY:
+            case PRIVATE_KEY:
                 signCallback.GotAuthorisation(true); //Legacy keys don't require authentication
```
James-Sangalli: @JamesSmartCell isn't this determined by whether the key is locked by the enclave rather than it's type? just because I import a private key, doesn't mean it can't be authenticated with the enclave

JamesSmartCell: No, legacy keys will never require authentication because these are legacy from before we allowed user to lock keys, and are handled by a different encryption algorithm.

Private key should never be seen here because that is only for import, once you import a private key it's converted into a KEYSTORE type for storage, so that switch statement will never trigger, however it will break signing with a key left over from the past.

A better fix would be to automatically convert all legacy keys to KEYSTORE type at startup which actually would simplify the code somewhat.

What is the actual issue you're seeing? When you import with a private key you should only be seeing that it's a yellow bar.

James-Sangalli: @JamesSmartCell I import one of @jzaki keys and I cannot make any transactions as it hits a problem in Authentication failed 

public enum AuthenticationFailType
{
    AUTHENTICATION_DIALOG_CANCELLED,
    FINGERPRINT_NOT_VALIDATED, PIN_FAILED,
    DEVICE_NOT_SECURE
}

We hit PIN_FAILED & DEVICE_NOT_SECURE both on the physical device and emulator, hence I thought it might be related...

JamesSmartCell: This is on a device that has no pin or lock?

JamesSmartCell: I just tested this: follow this flow on a non secured device:

- import private key
- attempt to move tokens or eth to another address

James-Sangalli:
Discussing on telegram

PR #1093, opened 2020-01-24
adds new updated strings for Chinese

James-Sangalli:
Will go downstream once merged


colourful-land:
I checked randomly a few - translations are correct but I found more than half of the phrases that I checked weren't used anywhere. We could have reduced our work by deleting these orphaned phrases?

Will @JamesSmartCell Please run the ["remove unused resource"](https://stackoverflow.com/questions/28937562/how-to-delete-an-unused-string-resource-for-all-configurations-in-android-studio) on this branch and push again?  I don't have Android Studio installed.

JamesSmartCell:
Yes some are orphaned like you say. Previously we neatly arranged and decluttered them but branch was lost. Needs to be done again on all 3 string sets.

PR #1095, opened 2020-01-24
Patch for mixed case address in keystore issue

JamesSmartCell:
Fix for #1098

Anomalous mixed case address in keystore was causing key issues (case sensitive file system, encoded using the address). Ensure address is stored and retrieved as a lower case file.

Note: Seed phrase keys are handled in the mixed case checksum format.


JamesSmartCell:
Did you reimport? You'll need to reimport.

colourful-land:
Your theory of doing "Ensure address is stored and retrieved as a lower case file." is that keystore has a HMAC check so there is no need for address lowercase/uppercase integrity check, right?

PR #1109, opened 2020-01-30
checksum wallet address in app

### LINKED ISSUE #445: Address is not checksumed
0xe188e8e7fca9e5d09b5a66148b072ecca936266b is my address on Android. everything is in lowercase. It remains so after being copied in to a clipboard.

Many services refuses to accept such Ethereum addresses with the concern that the user might have mistyped a character.

P.S. Consider storing checksummed address internally in the wallet (i.e. not only as a Big Integer) just in case.

James-Sangalli:
Fixes #445

Note: it still is lower case for contract address as this comes from Lib and lib cannot import web3J without breaking app code which also uses web3J


JamesSmartCell:
I don't understand the above comment. 

FYI we made a decision that checksum addr for legacy keystore was low priority. For Seed keys it's correct but doesn't validate on usage - which is a different set of code. It's not a good idea to try to bolt on checksum but it's ok to only display the address as checksum'd without changing the internal code.

PR #1111, opened 2020-01-30
fix wrong type and multiplication for eN multipliers

James-Sangalli:


James-Sangalli:
@JamesSmartCell A must before release as it breaks all the compound tokenscripts

PR #1129, opened 2020-02-03
uses TS name rather than contract name

### LINKED ISSUE #914: TokenScript cards should take the localised name defined in the TokenScript
At the moment it is taking the name from the smart contract, instead it should take the name defined in the tokenscript file and fallback to english if the user's language is not there. 

Useful for Devcon because we renamed the discovery card to devcon5 goodies but not a show stopper.

James-Sangalli:
Closes #914

Needs to go downstream


James-Sangalli:
Screen Shot 2020-02-03 at 2 19 46 pm

PR #1131, opened 2020-02-03
shows native tokens if set by user, even if the balance is zero

### LINKED ISSUE #904: Should show Zero balance native cards
For example, if I enable xDAI, I want to be able to see it there so I can add funds to it etc.

James-Sangalli:
Closes #904


James-Sangalli:
@JamesSmartCell I believe we should still show these, a user may want to send testnet tokens to themselves only to find it is not there.

JamesSmartCell:
The calue will appear once there's a positive balance. Otherwise we have a lot of zero cards. Other wallets don't have this issue with testnets as they don't show them.

It's probably ok since you can control visibility through the filters.

PR #1144, opened 2020-02-07
allows user added tokens to show up, even if balance is zero

James-Sangalli:
If a user adds a token, it should show up, even if the balance is zero. 

Testers are routinely confused by this and I would be too as a user and wanted to add a token manually.


JamesSmartCell:
Will be merged after Justin's token management patch is fully functional. This is the roadmap. 

Agree it's a UX fail but adding the token with zero balance and not giving users ability to remove the token as per this patch on its own would not be desirable - leading to a huge list of empty tokens. A better option is to show the balance as being zero on the token add screen and ask user if they want to display the empty token.

James-Sangalli:
Superseded by #1146

PR #1165, opened 2020-02-17
Update token cards more gracefully when there's a matching script

JamesSmartCell:
Fixes token update glitch seen when debug tokenscript matches address of token but not the chain - ie token with same address on different chains is created by same account and there's a script for one of those token but not the other. Rare but annoying.

Smooths out update when debug script is dropped in and closes a TODO.


James-Sangalli:
@JamesSmartCell so what happens now if the same address but different chain is triggered via a debug tokenscript?

JamesSmartCell:
It all works correctly. This is to fix an ugly update glitch that is very hard to get. I only saw it because I have two tokens on different chains with the same address, one of which has a script the other doesn't.

I also had a TODO: in the code that was causing the issue so it seemed like a good opportunity to resolve that at the same time. Works better now.

PR #1167, opened 2020-02-18
Dont allow zero balance TX

JamesSmartCell:
Warn when user trying to send a zero balance transaction.


James-Sangalli:
@JamesSmartCell you might have just rebased the definition of rebase ;)

PR #1170, opened 2020-02-18
Debounce TokenScript and Dapp signing

JamesSmartCell:
- Debounce TokenScript function signing.
- Debounce DappBrowser signing.
- Debounce TokenScript view click.
- Tidy up confirmation call.

Example issues that are fixed:

- Go to NFT factory. Set up token. Click 'deploy contract' button as fast as you can.
- Go to tokenscript entry token, go to 'enter' and then spam the confirm button as fast as you can. You can sometimes get the app into a 'zombie' state where the signing dialog doesn't do anything and can't be dismissed. It's because the real signing dialog is buried underneath.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/router/ConfirmationRouter.java
+++ b/app/src/main/java/com/alphawallet/app/router/ConfirmationRouter.java
@@ -36,14 +36,13 @@ public class ConfirmationRouter {
     }
 
     //Sign transaction for dapp browser
-    public void open(Activity context, Web3Transaction transaction, String networkName, boolean isMainNet, String requesterURL, int chainId) throws TransactionTooLargeException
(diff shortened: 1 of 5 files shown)


James-Sangalli:
@JamesSmartCell can you clearly write what the problem is and how the PR solved it.

JamesSmartCell:
The PR code fixes the button issues detailed above. It ensures you can't button mash the JavaScript web buttons and get multiple dialogs. Also during external testing we saw ppl pressing the send tx/sign button in between the first fingerprint check and the dialog popping up. This fix displays a progress spinner and removes focus from the button during that time, in all the places where this could possibly happen.

It's not an essential fix but helps reduce the chance of unexpected things happening during interaction.

We tested the same on iOS it's partly solved; we managed to get some weird things happening (unexpected dialogs popping up if you mash the 'deploy' JS button on token factory) but nothing that breaks the flow.

PR #1174, opened 2020-02-20
Handle Android 10 file system

JamesSmartCell:
Handle Android 10 style file intents and directories.

Previously, Android 10 implementation wasn't fully working for importing files.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -92,26 +92,26 @@
                     android:pathPrefix="/"
                     android:scheme="https" />
             </intent-filter>
-            <intent-filter>
-                <data
-                    android:host="*"
-                    android:mimeType="*/*"
-                    android:pathPattern=".*\\.tsml"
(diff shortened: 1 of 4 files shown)


James-Sangalli:
@JamesSmartCell could you please elaborate?

JamesSmartCell:
Android team at Google refactored the way external directories are used. Apps now have designated external (user accessible) directories, and shouldn't use 'custom' directories. If Apps need to store or access eg Pictures or Downloads they need to ask for that permission separately.

We saw that Android 10 users would encounter errors if they tried to use TokenScript with custom scripts going into the AlphaWallet directory.

This PR bridges the gap between pre-10 and Android 10 phones as originally we used the /AlphaWallet directory, which we're no longer allowed to use. It ensures legacy compatibility, newer phones with 10+ have to use the OS specified external directory - easy to get to dropping in scripts via intents (ie via Telegram or other chat packages) but slightly harder to actually find it for manual drop.

PR #1175, opened 2020-02-20
Fix for token select inconsistency with multiple token select.

JamesSmartCell:
- Fix for token select inconsistency with multiple token select. Only seen with multi-token select TokenScripts (eg cryptokitties). To reproduce issue: Use Cryptokitties breeding script, have multiple kitties; select a kitty, then select another kitty, deselect the first kitty, then transfer.

- Refactor and make it easier to add new tokens.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/widget/FunctionButtonBar.java
+++ b/app/src/main/java/com/alphawallet/app/widget/FunctionButtonBar.java
@@ -173,7 +173,8 @@ public class FunctionButtonBar extends LinearLayout implements OnTokenClickListe
             }
             else
             {
-                List<BigInteger> selected = adapter.getSelectedTokenIds(selection);
+                List<BigInteger> selected = selection;
+                if (adapter != null) selected = adapter.getSelectedTokenIds(selection);
 
                 switch (v.getId())
                 {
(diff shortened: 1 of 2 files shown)


JamesSmartCell:
This is for the multi token select. While testing it trying to do unexpected things I noticed it was possible to break the token selection as described above. If a bug is found, it should be fixed ... If the code can be improved so it deduplicates something (without breaking anything) while fixing then that should be done too. 

The above error reproduction shows how it's possible to break the selection system ...

JamesSmartCell:
To reproduce issue: Use Cryptokitties breeding script, have multiple kitties; select a kitty, then select another kitty, deselect the first kitty, then transfer. This will take you to a transfer screen with no kitty selected, and if you try to actually transfer that you get an incomplete confirmation screen which is clearly a UI fail. Reason: multi-select requires a tokenId 1 (long press) and tokenId 2 (select) since the order of tokenId in multi-select is important (it's the order they'll appear on the contract itself - eg breed kitty(token1, token2), token1 is the mother and token2 is the father. It came about as a result of extending the existing long-press transfer select.

It can only be seen on TokenScripts which have functions which have more than one tokenId as input. So far only the kitty breed script has that, so not many users would have been exposed to this issue.

While looking at the code I noticed some code function duplication which could result in issues with adding new token contract handling (eg ERC1155 etc) so I refactored to avoid that.

PR #1176, opened 2020-02-20
Explicitly switch off cached results for multi-tokenId input functions.

JamesSmartCell:
If a function has multi-token input (eg 'Kitty breed check) then don't use the cached result.


James-Sangalli:
@JamesSmartCell why?

JamesSmartCell:
https://community.tokenscript.org/t/volatile-tokenscript-attribute/241

Currently the cache doesn't handle multiple tokenId input - the long term solution is to adapt the cache to handle multiple tokenIds but a good short term fix is just to query the smart contract each time.

Since the kitties breeding script is the only thing that uses this it's not worth spending additional time on it. Devs may have more scripts with functions with multiple tokenIds - it's better that they work correctly than run fast.

PR #1191, opened 2020-02-25
Add Instagram

JamesSmartCell:
Added Instagram in settings page, should it be required.


James-Sangalli:
@JamesSmartCell we don't have an instagram?

PR #1201, opened 2020-02-28
Update Root Message

JamesSmartCell:
Updating the root warning, including Chinese string.

Hence - including all who can understand Mandarin.


JamesSmartCell:
Thanks, I used auto-translate and I think it didn't handle the second sentence very well

PR #1207, opened 2020-03-02
Formalise grouped token selection rule

JamesSmartCell:
Formalise grouping rule for tokens. Since the token contract determines if we can transfer multiple tokens in a single transfer it makes sense to add this property to the token contract definition.

This will make it easier to add new NFT's - you just specify if it has multiple token send and the group selection will function accordingly.

Tested with ERC875, 721 and 721T.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokens/Ticket.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokens/Ticket.java
@@ -513,6 +513,8 @@ public class Ticket extends Token implements Parcelable
     @Override
     public boolean isERC875() { return true; }
     public boolean isNonFungible() { return true; }
+    @Override
+    public boolean hasGroupedTransfer() { return true; }
 
     @Override
     public boolean groupWithToken(TicketRange currentGroupingRange, TicketRangeElement newElement, long currentGroupTime)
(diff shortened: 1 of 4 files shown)


James-Sangalli:
@JamesSmartCell cannot select multiple tokens, could you explain how to reproduce this?

JamesSmartCell:
It needs to be declared in the TokenScript. Android picks up Ethereum functions with multiple tokenId input declared in the attributes and knows you need multiple token select based on that. Do you have cryptokitties Rinkeby? If not I'll breed some for you. I added the cryptokitties breeding script into the Tokenscript examples repo for you to try it.

The script should also work with Main-net kitties but I haven't tested it with those yet.

James-Sangalli:
@JamesSmartCell this is a schema change. @colourful-land I'll let you handle this

colourful-land:
> @JamesSmartCell this is a schema change. @colourful-land I'll let you handle this

I'm on it.

JamesSmartCell:
We actually need this PR regardless. It's not a schema change, it's a formalisation of when there's multi-token selection, which is something that will happen. The actual schema change affects a different part of the app; in the interpretation area. This is the UX rendering area which receives those instructions.

colourful-land:
> We actually need this PR regardless. It's not a schema change, it's a formalisation of the UI for when there is multi-token selection, which is something that will happen. The actual schema change affects a different part of the app; in the schema parsing area.

Look like @JamesSmartCell is right, that there is no schema change or anticipated changes in this PR>

Issue #1136, opened 2020-02-04
Add Localised currency


JamesSmartCell:
This is a mini project which lets users switch their local currency eg AUD, GBP, INR, CNY etc.

It will need: 
- new item in settings for major currencies (at minimum add GBP, AUD, INR, EUR and CNY).
- Create new display class to show currency which automatically converts currency.
- Replace all instances of currency display and usage to use new class.

There is an API in 'TickerService' which converts currencies. All fiat currency values are received in USD; this shouldn't change - just apply local conversion in the new class.

ChintanRathod:
![image](https://user-images.githubusercontent.com/4371780/73836508-6e58c780-4835-11ea-9454-fcefde03bc00.png)

For now implemented necessary currency list into settings. 
@tomekalphawallet Kindly review the design if we need to apply some changes for this issue.

tomekalphawallet:
@ChintanRathod This is a great start. Please keep in mind that we have a new design system. So as long you tried to mimic the current design, really soon we will have a new one, where all fonts, colors, input types are aligned. 

Note: 

a) I added "FIAT" to the title. 
b) We won't have a Confirm button. You confirm by hitting the back button. 
c) I updated the design to be aligned with the new design system (we use checkmark for multiple choices)

Do you have an access to Zeplin libraries? 

Updated design:
![Select_FIAT_Currency](https://user-images.githubusercontent.com/51817359/73918571-28961080-48c2-11ea-99cf-a2b441de3bec.png)

JamesSmartCell:
[flags_svg.zip](https://github.com/AlphaWallet/alpha-wallet-android/files/4301269/flags_svg.zip)

@ChintanRathod here are the round flag SVG assets from the above screen

JamesSmartCell:
Complete

PR #1210, opened 2020-03-05
Implement Task #1136

ChintanRathod:
- Add 6 (USD, EUR, GBP, AUD, CNY, INR) currencies at initial level to select from setting screen
- Add initial selection of USD and later user can select own default currency from the list
- Stored selection to local preference
- Change TickerService to apply changed currency to wallet display


JamesSmartCell:
Hi @ChintanRathod . This looks great.

Could I ask for just two small mods though.

First - the currency icons should be round; can you use the ones that @tomekalphawallet posted above - you may need to tweak them when you import but I think the import feature on Android Studio generates rounded icons.

Second - your resume method in HomeActivity - can it call the progress spinner in WalletFragment so the user knows something is being updated (```systemView.showProgress(true);```)? Then, when the ticker update completes the currency update and calls refreshTokens() in WalletFragment it can switch the progressSpinner off ```systemView.showProgress(false);```

ChintanRathod:
@JamesSmartCell Thanks for sparing time to review the PR and sharing feedback. I will check and update the code accordingly. I will check for the rounded icons. If I faced any issue, will let you know to have some help from @tomekalphawallet.

ChintanRathod:
@JamesSmartCell As per instructed, implemented first modification of circular icon with newer design proposed by @tomekalphawallet. The second modification was not able to implemented due to some new widget integration into home activity.
Would you please review the additional changes and let me know if need any change.

JamesSmartCell:
Great, we can now add new currencies in a few seconds.
The spinner issue wasn't as clear cut as I thought due to some legacy code confusing the issue. I tidied this up.

PR #1220, opened 2020-03-11
Streamlined Dappbrowser (no URL) option

JamesSmartCell:
Implement minimal Dappbrowser appearance selectable from VisibilityFilter.


app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
```diff
@@ -429,44 +433,55 @@ public void onDestroy()
...
-                    viewModel.startScan(getActivity());
-                    return true;
-                });
-        toolbar.getMenu().findItem(R.id.action_add_to_my_dapps)
-                .setOnMenuItemClickListener(menuItem -> {
-                    viewModel.addToMyDapps(getContext(), currentWebpageTitle, urlTv.getText().toString());
-                    return true;
-                });
+        if (VisibilityFilter.minimiseBrowserURLBar()) toolbar.inflateMenu(R.menu.menu_scan);
+        else toolbar.inflateMenu(R.menu.menu_bookmarks);
```
jzaki: `menu_dapp_browser` instead of `menu_bookmarks` since it's not a menu about bookmarks.

app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
```diff
@@ -429,44 +433,55 @@ public void onDestroy()
         super.onDestroy();
     }
 
+    private void setupMenu(View baseView)
+    {
+        refresh = baseView.findViewById(R.id.refresh);
```
jzaki: `refresh` set before, and only used after, being set in this function.

app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
```diff
@@ -164,6 +166,7 @@
     private ImageView back;
     private ImageView next;
     private ImageView clear;
+    private ImageView refresh;
```
jzaki: Since this is the reload image view, recommend naming, `reload`, for consistency.

jzaki: Or are refresh and reload different things here?

app/src/awallet/java/com/alphawallet/app/entity/VisibilityFilter.java
```diff
@@ -98,4 +98,7 @@ public static boolean canChangeWallets()
 
     //In main wallet menu, if wallet allows adding new tokens
     public static boolean canAddTokens() { return true; }
+
+    //Implement minimal dappbrowser with no URL bar.
+    public static boolean minimiseBrowserURLBar() { return false; }
```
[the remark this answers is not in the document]
JamesSmartCell: This is a command function; ```hasVisibleBrowserURLBar``` erroneously indicates this is a variable getter from the method.

The functions here determine how the app appears, rather than fetching info about how they are. Maybe shouldMinimiseBrowserURLBar() would be better?

app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
JamesSmartCell: Not sure let's follow brave/chrome - which use 'reload' so you're right.

Issue #722, opened 2019-06-18
Provide a Clear Browser Cache function in the setting page


zhangzhongnan928:
https://github.com/AlphaWallet/alpha-wallet-ios/issues/1244

ChintanRathod:
Started working on this issue, yet stuck at below points. 
1) Image for clear browser 
2) Translation. What is your approach for translations? If it is Google translations, I can do that, else I need translated strings from relevant person.

ronakdave195:
@James-Sangalli We await to receive your feedback on @ChintanRathod's above query

justindg:
Hi @ronakdave195  and @ChintanRathod , I'm currently working on the revamped Settings page. The new page will include the section for the "Clear Browser Cache" function, so no need to worry about the layout and icon. I will leave a corresponding TODO tag in the source for you to write the actual implementation of the cache clear function. I will update this ticket once the PR is up. :)

justindg:
Update: You can now add the implementation in **`AdvancedSettingsActivity`** under `onClearBrowserCacheClicked()`

ChintanRathod:
@justindg Sure. I will add the logic in AdvancedSettingsActivity and push the code in "Fiji" branch.

PR #1232, opened 2020-03-15
Implement Task #722

ChintanRathod:
- Add code to clean browser cache and show message to the user

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/AdvancedSettingsActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/AdvancedSettingsActivity.java
@@ -9,7 +9,9 @@ import android.support.annotation.Nullable;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.ContextCompat;
 import android.util.Log;
+import android.webkit.WebView;
 import android.widget.LinearLayout;
+import android.widget.Toast;
 
 import com.alphawallet.app.C;
 import com.alphawallet.app.R;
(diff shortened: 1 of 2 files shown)


justindg:
Hi @ChintanRathod, I think the requirement is to clear the cache of the Web3View in DappBrowserFragment? 

@tomekalphawallet would you require a dialog for user confirmation before clearing the cache?

ChintanRathod:
@justindg I understand, but Web3View is ultimately using WebView as base and it has all implementation. Please check [this link](https://developer.android.com/reference/android/webkit/WebView.html#clearCache%28boolean%29) which defines if WebView cache directory is wiped, it's associated all instance data will be wiped because of single location cache mechanism used by Android OS.

JamesSmartCell:
I'll have to check it but if it works it's a clever solution!

JamesSmartCell:
@ChintanRathod please add me as reviewer on pull requests

justindg:
> @justindg I understand, but Web3View is ultimately using WebView as base and it has all implementation. Please check [this link](https://developer.android.com/reference/android/webkit/WebView.html#clearCache%28boolean%29) which defines if WebView cache directory is wiped, it's associated all instance data will be wiped because of single location cache mechanism used by Android OS.
> 
> As defined in supporting document of Google, it is not necessary to clear the cache at used instance. We can wipe anywhere in the application as cache used by WebView is common per application basis.

Never knew this, this is brilliant!

ChintanRathod:
@JamesSmartCell, please review this. 
@justindg Thank you :)

PR #1234, opened 2020-03-16
Fix asset instance race condition

JamesSmartCell:
Fixes a race condition only seen when restarting the app while a Token Card is visible.

After the wallet's local memory is reclaimed by the OS and the wallet is restarted the asset service re-loads the scripts.

At the same time if the wallet was viewing a token card, the view will check if the token has an asset.

This PR introduces a semaphore to ensure the token card query holds until loading is complete.

A subsequent PR will store the token -> contract mappings into a database for rapid reload.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
@@ -8,6 +8,7 @@ import android.text.TextUtils;
 import android.util.Base64;
 import android.view.View;
 import android.widget.ProgressBar;
+import android.widget.TextView;
 
 import com.alphawallet.app.R;
 import com.alphawallet.app.entity.ContractType;
@@ -24,6 +25,7 @@ import com.alphawallet.app.viewmodel.BaseViewModel;
 import com.alphawallet.app.web3.Web3TokenView;
 import com.alphawallet.token.entity.TicketRange;
 import com.alphawallet.token.entity.TokenScriptResult;
+import com.alphawallet.token.tools.TokenDefinition;
 
 import org.web3j.abi.datatypes.Function;
 import org.web3j.utils.Numeric;
@@ -36,6 +38,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -604,11 +607,6 @@ public class Token implements Parcelable, Comparable<Token>
     public void checkIsMatchedInXML(AssetDefinitionService assetService) { }
     public int[] getTicketIndices(String ticketIds) { return new int[0]; }
     public boolean contractTypeValid() { return !(contractType == ContractType.NOT_SET || contractType == ContractType.OTHER); }
-
-    public void displayTicketHolder(TicketRange range, View activity, AssetDefinitionService assetService, Context ctx, boolean iconified) { }
-    public void displayTicketHolder(TicketRange range, View activity, AssetDefinitionService assetService, Context ctx) {
-        displayTicketHolder(range, activity, assetService, ctx, false);
-    }
     public List<BigInteger> getArrayBalance() { return new ArrayList<>(); }
     public List<BigInteger> getNonZeroArrayBalance() { return new ArrayList<>(Arrays.asList(BigInteger.ZERO)); }
     public boolean isMatchedInXML() { return false; }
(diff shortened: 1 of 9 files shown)


JamesSmartCell:
> It looks working to me, but I have a question:
> Why using Rx here to spawn threads to read each XML file? Why not do it in a linear fashion?

It makes the code cleaner - and once added to database would only pick up new or changed files.

app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
```diff
@@ -346,12 +336,26 @@ public TokenDefinition getAssetDefinition(int chainId, String address)
...
+    public Single<TokenDefinition> getAssetDefinitionASync(int chainId, final String address)
     {
         if (address == null) return Single.fromCallable(TokenDefinition::new);
         String contractName = address;
         if (contractName.equalsIgnoreCase(tokensService.getCurrentAddress())) contractName = "ethereum";
 
+        // hold until asset definitions have finished loading
+        try
+        {
+            assetLoadingLock.acquire();
```
jzaki: If an InterruptedException occurs, the semaphore is not acquired, so the `release` in finally is wrong.
`release` should be right after acquire() in this usage.

app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
```diff
@@ -781,20 +819,25 @@ private void checkDownloadedFiles()
                 .subscribe(this::handleFileLoad, this::onError).isDisposed();
     }
 
-    private Single<File> updateSignature(File file)
+    /* Add cached signature if uncached files found. */
+    private Single<File> cacheSignature(File file)
     {
+        // note that outdated cache is never deleted - we don't have that level of finesse
```
jzaki: Recommend capturing memory bloat in an issue and referencing here

app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
```diff
@@ -1433,17 +1474,6 @@ public boolean viewsEqual(Token token)
...
-     * If the OS has scavenged the TokenScript contract lookup map then re-init
-     */
-    private void reloadAssets()
-    {
-        if (assetDefinitions == null || assetDefinitions.size() == 0)
-        {
-            loadLocalContracts();
-        }
-    }
-
```
jzaki: File is over 1500 lines, possible to divide

app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
colourful-land: the cache is not in the memory btw. they are on mobile storage.

app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
```diff
@@ -240,32 +254,9 @@ public TokenScriptResult getTokenScriptResult(Token token, BigInteger tokenId)
...
-                            + File.separator + HomeViewModel.ALPHAWALLET_DIR);
-
-            if (!directory.exists())
-            {
-                directory.mkdir(); //does this throw if we haven't given permission?
-            }
-
-            loadExternalContracts(directory, true);
-        }
+    private boolean checkWritePermission() {
```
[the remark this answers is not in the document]
colourful-land: this one is already merged so you might start a PR to rename it. Also that naming rule isn't universally true. `hasXXX` implies the action not taking resources and non-blocking (e.g. no need to wait) while `checkXXX` means the action takes resource (e.g. don't call it form a UI thread) or might block. e.g. if there is an RESTful API call, it's `checkXXX`.

app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
```diff
@@ -1022,6 +1020,55 @@ public boolean groupWithToken(TicketRange currentRange, TicketRangeElement e, lo
...
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(td -> renderTicketHolder(td, range, activity, assetService, ctx, iconified), this::loadingError).isDisposed();
+    }
+
+    private void loadingError(Throwable e)
+    {
+        e.printStackTrace();
+    }
+
+    private void renderTicketHolder(TokenDefinition td, TicketRange range, View activity, AssetDefinitionService assetService, Context ctx, boolean iconified)
```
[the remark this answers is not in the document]
colourful-land: Ticket might be moved away from Token entirely since it might be addressing Ticket-like token (FIFA).  @JamesSmartCell share some knowledge here?

PR #1235, opened 2020-03-16
Spotfix for Currency update

JamesSmartCell:
Hi @ChintanRathod here's a small spotfix to get the ticker update working. It was pretty much there, just needed to poke the tickerService to start the update process then receive the completion from the receiver (which you already had).


ChintanRathod:
@JamesSmartCell It works as expected, thanks for fixing :)

PR #1238, opened 2020-03-16
Fixes  #1236

### LINKED ISSUE #1236: Improve Flow of back button
Most Android apps have a back button guard, where if you have multi-fragment app they return you to the 'head' page of the fragments if you click back. Either they will make you press back twice or return you to the 'home' page.

Currently, if you're on any of the 3 main fragment pages other than the default wallet page (Transactions, Settings or Browser) and you click back you leave the app, which in most cases is unexpected.

Update this behaviour so that a back click when on Transactions, Browser or Settings will take you to the wallet page.

A back click on the wallet page will leave the app.

See also #781, #800, #787

Implement this on the Fiji-Implementation branch.

ChintanRathod:
Fixes #1236

- Check for current showing item on bottom navigation
- Once user press back button if not in WALLET page, it will navigate to WALLET fragment else close the app

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/HomeActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/HomeActivity.java
@@ -949,4 +949,17 @@ public class HomeActivity extends BaseNavigationActivity implements View.OnClick
         sendBroadcast(new Intent(CHANGED_LOCALE));
         viewModel.updateLocale(newLocale, this);
     }
+
+    @Override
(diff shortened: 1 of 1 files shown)


JamesSmartCell:
Hi @ChintanRathod , thanks for this I will check it out shortly.

For future pull requests can you give the PR a name which is a summary of the PR, eg for this one you could just use the Issue title like 'Improve flow of back button'. This makes it easier to index the commit history. Thanks :)

PR #1241, opened 2020-03-18
adding comment to clarify the use of semaphore

colourful-land:


app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
```diff
@@ -133,6 +134,11 @@ public AssetDefinitionService(OkHttpClient client, Context ctx, NotificationServ
         loadLocalContracts();
     }
 
+    /* Spawns a few threads to load contracts, then spawns more, than
```
JamesSmartCell: This comment should be a PR comment not a code comment. The comment here doesn't help with understanding what is being done or why.

There is scope for refactoring. There are better ways of doing this once Android 23-25 no longer need to be supported.

I would say - if you consider there is a need to write what the semaphore is used for -

1. Lock the semaphore.
Parse the three sets of tokenscripts to create the mapping between chainId/address and script file  -
2. Downloaded from repo.
3. Installed (developer) tokenscripts.
4. Installed (developer) tokenscripts in legacy directory.

Once complete release the semaphore lock.

The code to render the token cards will be locked by the semaphore until loading is complete.

PR #1250, opened 2020-03-22
Update Function Bar

### LINKED ISSUE #1149: Update TS Buttons
Sketch Cloud:
- https://sketch.cloud/s/jDODG/a/nn1YDz
- https://sketch.cloud/s/jDODG/a/QG9g17
- https://sketch.cloud/s/jDODG/a/xn98Qv
- https://sketch.cloud/s/jDODG/a/EMo1G2
- https://sketch.cloud/s/jDODG/a/jElRrz

and few more if you switch to next to the right.. (Function 3,4,5,6)

Zeplin:
https://app.zeplin.io/project/5d088205bff2d15de6a4397b/dashboard?seid=5e41333605a98088a65653b4

justindg:
Closes #1149 

Summary of changes:
- FunctionBar can now be inflated via XML as a View, instead of using include
- FunctionBar has 3 fixed components: **Primary Button**, **Secondary Button**, and **Menu/More Button**
- Each component's functionality will still be assigned via `setupFunctions`
- The functions are now mapped via the `String` value of the resource. Previously it was either a `String` or `int`. Opted to use `String` exclusively for simplicity.
- For generic functions, `handleClick()` now accepts a `String` instead of `int` as parameter
- The **third** and succeeding functions will be added to a ListView, and will be accessed via the **Menu/More Button**

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/StandardFunctionInterface.java
+++ b/app/src/main/java/com/alphawallet/app/entity/StandardFunctionInterface.java
@@ -13,6 +13,6 @@ public interface StandardFunctionInterface
     default void showSend() { };
     default void showReceive() { };
     default void displayTokenSelectionError(TSAction action) { };
-    default void handleClick(int view) { };
+    default void handleClick(String action) { };
     default void handleTokenScriptFunction(String function, List<BigInteger> selection) { };
 }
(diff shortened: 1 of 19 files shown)


JamesSmartCell:
@justindg made a small tweak to tidy up my old code and also change the button order for Non Fungibles vs Fungibles. Generally send/receive are low priority for Non Fungibles - which tend be used for actions more than transfer, whereas fungibles tend to be more used for transfer.

PR #1256, opened 2020-03-25
add MIT license to our modules

James-Sangalli:
Since these modules are separate from the android project itself, they should be licensed by us with the MIT license.


hboon:
@James-Sangalli this will provide more freedom for everyone (not just white-labels) including for closed-source products. Is that the intention?

James-Sangalli:
> @James-Sangalli this will provide more freedom for everyone (not just white-labels) including for closed-source products. Is that the intention?

Yes

PR #1260, opened 2020-03-31
Update JavaScript API

### LINKED ISSUE #1185: adding JavaScript API (android)
Incorporate new javascript object interpretation, example: 

`web3.tokens.dataChanged = (oldTokens, updatedTokens, tokenCardId) => {
    const currentTokenInstance = updatedTokens.currentInstance;
    document.getElementById(tokenCardId).getElementsByClassName("contents")[0].innerHTML = new Token(currentTokenInstance).render()
};`

JamesSmartCell:
Fixes #1185 

- Supply tokenId to view to create the token-card wrapper.
- Allow legacy scripts with 'view-iconified' to still render correctly.

TODO in additional PR:
- Cache the injected ```_currentTokenInstance``` to use as the first param in ```dataChanged```.
- Mark 'view-iconified' as deprecated in the TokenScript management page.

### DIFF
--- a/lib/src/main/java/com/alphawallet/token/tools/TokenDefinition.java
+++ b/lib/src/main/java/com/alphawallet/token/tools/TokenDefinition.java
@@ -894,39 +894,21 @@ public class TokenDefinition {
     }
 
     /**
-     * Check for 'appearance' attribute set
-     * @param tag
-     * @return
-     */
-    public String getAppearance(String tag)
-    {
-        Map<String, String> appearanceSet = attributeSets.get("appearance");
-        if (appearanceSet != null)
-        {
-            return appearanceSet.get(tag);
-        }
-        else
-        {
-            return "";
-        }
-    }
-
-    /**
      * Check for 'cards' attribute set
      * @param tag
      * @return
      */
     public String getCardData(String tag)
     {
+        String view = null;
         Map<String, String> appearanceSet = attributeSets.get("cards");
         if (appearanceSet != null)
         {
-            return appearanceSet.get(tag);
+            view = appearanceSet.get(tag);
+            if (view == null && tag.equals("item-view")) view = getCardData("view-iconified"); //deal with legacy views
         }
-        else
-        {
-            return "";
-        }
+
+        return view != null ? view : "";
     }
 
     public Map<String, TSAction> getActions()
--- a/lib/src/main/java/com/alphawallet/token/tools/TokenDefinition.java
+++ b/lib/src/main/java/com/alphawallet/token/tools/TokenDefinition.java
@@ -38,6 +38,9 @@ public class TokenDefinition {
     private static final String TOKENSCRIPT_NAMESPACE = "http://tokenscript.org/" + TOKENSCRIPT_CURRENT_SCHEMA + "/tokenscript";
     private static final String TOKENSCRIPT_BASE_URL = "http://tokenscript.org/";
 
+    public static final String TOKENSCRIPT_ERROR = "<h2 style=\"color:rgba(207, 0, 15, 1);\">TokenScript Error</h2>";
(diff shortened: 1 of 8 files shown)


lib/src/main/java/com/alphawallet/token/tools/TokenDefinition.java
```diff
@@ -893,40 +893,22 @@ public void parseField(BigInteger tokenId, NonFungibleToken token) {
...
         Map<String, String> appearanceSet = attributeSets.get("cards");
         if (appearanceSet != null)
         {
-            return appearanceSet.get(tag);
-        }
-        else
-        {
-            return "";
+            view = appearanceSet.get(tag);
+            if (view == null && tag.equals("item-view")) view = getCardData("view-iconified"); //deal with legacy views
```
James-Sangalli: @JamesSmartCell we can probably just drop the old way, best to leave behind all the legacy stuff and move on

JamesSmartCell: It's may not be a good idea to just drop something like that - no language I know ever phased out anything abruptly - you mark it deprecated then remove a few months later. Depends how many people wrote scripts already. If they couldn't get them working because of this ...

JamesSmartCell: Here's how it will work: let's remove the legacy handling but add an error message to show what happened should anyone ever come across this.

James-Sangalli: @JamesSmartCell I think this point is moot because we don't support legacy versions anyway and the new schema will not validate if you use the old way; therefore supporting the old syntax doesn't add anything. Regarding phasing out, swift does this all the time, just ask @hboon. 

Also remember that there are few people writing them right now so we might as well eliminate the older versions while we still can.

What do you think?

James-Sangalli: @colourful-land

James-Sangalli: @JamesSmartCell made the point that for future proofing we should add a warning that it is old, this is a good idea but a different PR.

JamesSmartCell: @James-Sangalli Removed legacy support, added error feedback as per suggestion from @colourful-land to show the view is deprecated and link to schema description.

lib/src/main/java/com/alphawallet/token/tools/TokenDefinition.java
```diff
@@ -905,12 +908,35 @@ public String getCardData(String tag)
...
+    private String checkLegacyCards(String tag, Map<String, String> appearanceSet)
+    {
+        String legacyWarningMessage = null;
+
+        switch (tag)
+        {
+            case "item-view":
+                if (appearanceSet.containsKey("view-iconified"))
+                {
+                    legacyWarningMessage = LEGACY_WARNING_TEMPLATE.replace("${ERR1}", "view-iconified")
```
James-Sangalli, line 926: @JamesSmartCell better as if statement?

colourful-land: If you are pretty sure just do a commit aggressively and wait for @JamesSmartCell to blame you.

lib/src/main/java/com/alphawallet/token/tools/TokenDefinition.java
colourful-land: Old script will still work because old script provides its own wrapper (adding one doesn't change anything) and the reactJS update call is designed (by @hboon) to be backward compatible. That's why I asked this to be done first, before even forking upgrade-schema.

lib/src/main/java/com/alphawallet/token/tools/TokenDefinition.java
JamesSmartCell: Prefer switch statements as clearer to follow.

PR #1261, opened 2020-03-31
Normalise TokenScript functionality

JamesSmartCell:
The PR normalises the response of the App to TokenScript across all the views and bundles TokenScript response code into a single source (The TokenFunctionViewModel + FunctionButtonBar combo).

While getting the event scripts working it was noticed that TokenScript behaviour was different across all activities that use and render TokenScript:

- AssetDisplayActivity - where you land when you click on a token in the wallet - displays list of rendered tokenId - ```<item-view>``` (previously ```<view-iconified>```)
- TokenFunctionActivity - where you land if you click on a ERC721Ticket/ERC875 token in AssetDisplayActivity - displays individual or grouped range of tokenIds - uses ```<view>```
- FunctionActivity - where you land if you click on a TokenScript ```<action>```
- TokenDetailActivity - where you land when you click on an ERC721

This PR normalises the app response from any TokenScript Action or Card seen on any page. For example - if a script has a cardless action (eg send Ethereum transaction) previously this only worked from FunctionActivity, ie an ```<action>```. This PR ensures it works if placed within a ```<view>```.

Also, previously TokenScript actions weren't implemented for pure ERC721 token functions.

Post this PR there's a consistent response to TokenScript no matter where the view is in the app.
It also is de-duplicates a lot of TokenScript handling.


James-Sangalli:
@JamesSmartCell looks good to me but how should I test?

JamesSmartCell:
Have a script for an NFT with a card view with an <action> that has no card, ie pure eth send. Use a normal ERC875 script and add this action: 

```
<ts:action>
      <ts:name>
        <ts:string xml:lang="en">Poke</ts:string>
      </ts:name>
	    <ts:transaction>
            <ts:ethereum as="bool" contract="<Your origin contract>" function="testFunc">
                <ts:data>
                    <ts:uint256 ref="tokenId"/>
                </ts:data>
            </ts:ethereum>
        </ts:transaction>
    </ts:action>
```
The function will fail if you actually try to push it as there's no 'testFunc' in the contract but it should take you from the item-view page (the first one after you click on the token in the wallet) straight to the transaction confirmation page.

Alternatively if you have any Rinkeby cryptokitties (or you point the origin at mainnet kitties) you can use the kitties script I posted in examples and 'breed' from the ERC721 details page (where you see all the extended attributes).

These two previously actions wouldn't have previously worked from these places.

colourful-land:
Not holding up this PR but TokenFunctionModule class sounds like a bag class (bag class = everything can be put there, the "Class is a container of similar functions" thinking that was once popular in the Java BootCamp teaching).

colourful-land:
I can't review this. This requires too much working memory to check. Would be a lot better as 6 separate PRs. Anyway, it look like JB cleaning up coding shortcuts that was made in the past year, so I have no problem having it merged in.

JamesSmartCell:
The PR is not just 'fixing viewless class', that is one of the symptoms of the underlying issue. I wrote a detailed report of what this PR does, and why. I'm not sure how much extra detail is needed to add to the description, or whether that would have any value since it's not being read/understood.

Refactor is big part of software lifecycle especially when following moving specs. Once the specs solidify almost always you can re-visit the code and improve as the overall design has become more clear and the problem understood. 

It would be trivial, when the time comes to improve the code to point fingers and ask, in hindsight why the improvement is necessary. But of course, we would never do that as we're professionals who understand how software is written.

Maybe looking at the diff alone it's difficult to see what's happening.

Issue #1194, opened 2020-02-26
Add ability to retrieve events inside TokenScript


James-Sangalli:
https://gist.github.com/James-Sangalli/0509258f3d56b66b68bfd6d51a10639a

James-Sangalli:
@JamesSmartCell I'll add the function and a test to retrieve the events and then you can integrate it in

James-Sangalli:
@JamesSmartCell as discussed, here is a code snippet for retrieving filtered events in Web3J with sample output: https://gist.github.com/James-Sangalli/4d1f031d5cf6ac724e7946b48f1cf44c

JamesSmartCell:
Can now access events.

Issue #1180, opened 2020-02-24
Android: interpretation a filter (currently only key=value)


James-Sangalli:
@colourful-land said this:

I envision the final product is a filter that can be used to obtain object in:

1) attestations, especially when one attestation depends on another with specific predicate (which is a filter)
2) tokens, during token negotiation
3) data objects, e.g. finding events.

But in this issue, we only need a filter that can obtain an event through a key=value format. That is, αWallet can parse the filter used in line 139

https://github.com/AlphaWallet/TokenScript-Examples/blob/27ffe6a83a7c41631ebfe22e262e21850f234327/examples/ENS/ENS.xml#L137-L141

PR #1262, opened 2020-03-31
Add Event Parsing

### LINKED ISSUE #1183: Android adding ASNX module parsing capacity (only used for events now)
It means iOS should be aware of this file's line 26 to line 34

https://github.com/AlphaWallet/TokenScript-Examples/blob/27ffe6a83a7c41631ebfe22e262e21850f234327/examples/ENS/ENS.xml#L26-L34

JamesSmartCell:
Add event parsing and some limited error reporting. 

After event parsing is complete the error handling will be completed so any tokenscript errors can be immediately seen when script is dropped into the app folder.

Fixes #1183
Adds instrumentation to allow operation of #1180 and #1194

### DIFF
--- a/lib/src/main/java/com/alphawallet/token/entity/AttributeType.java
+++ b/lib/src/main/java/com/alphawallet/token/entity/AttributeType.java
@@ -34,12 +34,14 @@ public class AttributeType {
     public Map<BigInteger, String> members;
     private TokenDefinition definition;
     public FunctionDefinition function = null;
+    public EventDefinition event = null;
     public boolean userInput = false;
 
     public AttributeType(Element attr, TokenDefinition def)
     {
         definition = def;
         id = attr.getAttribute("id");
+        name = id; //set name to id if not specified
         as = As.Unsigned; //default value
         try {
             switch (attr.getAttribute("syntax")) { // We don't validate syntax here; schema does it.
@@ -66,6 +68,9 @@ public class AttributeType {
                     break;
                 case "1.3.6.1.4.1.1466.115.121.1.27":
                     syntax = TokenDefinition.Syntax.Integer;
+                    break;
+                case "1.3.6.1.4.1.1466.115.121.1.15":
+                    syntax = TokenDefinition.Syntax.DirectoryString;
                     break;
                 default: // unknown syntax treat as Directory String
                     syntax = TokenDefinition.Syntax.DirectoryString;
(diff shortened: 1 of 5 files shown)


lib/src/main/java/com/alphawallet/token/entity/AttributeType.java
```diff
@@ -234,6 +247,30 @@ else if (data.startsWith("0x"))
         }
     }
 
+    //Sometimes value needs to be processed from the raw input.
+    //Currently only time
+    public BigInteger processValue(BigInteger val) throws UnsupportedEncodingException, ParseException
+    {
+        switch (syntax)
+        {
+            case GeneralizedTime:
```
James-Sangalli, line 256: @JamesSmartCell want to add the missing cases? if not implemented we can still put them in and simply break

lib/src/main/java/com/alphawallet/token/entity/ContractInfo.java
```diff
@@ -12,4 +12,12 @@
 {
     public String contractInterface = null;
     public Map<Integer, List<String>> addresses = new HashMap<>();
+    public Map<String, Module> eventModules = null;
+
+    public boolean hasContract(int chainId, String address)
```
James-Sangalli, line 17: @JamesSmartCell could this be named for specifically?

lib/src/main/java/com/alphawallet/token/entity/AttributeType.java
```diff
@@ -67,6 +69,9 @@ public AttributeType(Element attr, TokenDefinition def)
                 case "1.3.6.1.4.1.1466.115.121.1.27":
```
James-Sangalli, line 69: Could we define these as constants to make them less error prone and simple?

lib/src/main/java/com/alphawallet/token/entity/AttributeType.java
JamesSmartCell: There's no missing cases. Currently we only need to transform time attribute value (ie convert the string representation to Unix time value). In future there may be other attributes that need transforming, but currently we don't know what they are.

lib/src/main/java/com/alphawallet/token/entity/ContractInfo.java
JamesSmartCell: which? eventModules or hasContract?

lib/src/main/java/com/alphawallet/token/entity/AttributeType.java
JamesSmartCell: those are the ISO values which are used in the tokenscript. Are you proposing something like an enum abstraction? If so it's not a bad idea.

lib/src/main/java/com/alphawallet/token/entity/ContractInfo.java
James-Sangalli: hasContract means has tokenscript for contract?

lib/src/main/java/com/alphawallet/token/entity/AttributeType.java
James-Sangalli: Added the missing but not used cases like Boolean, BitString etc. While these are not being implemented, the switch should be exhaustive otherwise it would be better to just use an if statement

lib/src/main/java/com/alphawallet/token/entity/AttributeType.java
James-Sangalli: Done

PR #1263, opened 2020-04-01
Add Action local attributes

JamesSmartCell:
Adds the local attributes for an ```<action>``` into the injected token properties.

Allows code like this to work (taken from ENS script):

```
<ts:action>
      <ts:name>
        <ts:string xml:lang="en">Renew</ts:string>
      </ts:name>

      <ts:attribute-type id="renewalPricePerYear" syntax="1.3.6.1.4.1.1466.115.121.1.36">
        <ts:name>
          <ts:string xml:lang="en">renewal price per year</ts:string>
        </ts:name>
        <ts:origins>
          <ts:ethereum function="rentPrice" contract="ETHRegistrarController" as="uint">
            <ts:data>
              <ts:string ref="ensName"/>
              <ts:uint256>31540000</ts:uint256>
            </ts:data>
          </ts:ethereum>
        </ts:origins>
      </ts:attribute-type>
```
...
```
<h3>Price of one year renewal: ${(this.props.renewalPricePerYear / 1e+18).toFixed(3)} ETH</h3>
```

The renewalPricePerYear attribute is only injected into the 'renew' action card view, not into any other actions or views.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
@@ -1089,7 +1089,7 @@ public class Token implements Parcelable, Comparable<Token>
 
         final StringBuilder attrs = assetService.getTokenAttrs(this, tokenId, range.tokenIds.size());
 
-        assetService.resolveAttrs(this, tokenId)
+        assetService.resolveAttrs(this, tokenId, null) //Can a view have local attributes?
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
```diff
@@ -1089,7 +1089,7 @@ public void renderTokenscriptView(TicketRange range, AssetDefinitionService asse
 
         final StringBuilder attrs = assetService.getTokenAttrs(this, tokenId, range.tokenIds.size());
 
-        assetService.resolveAttrs(this, tokenId)
+        assetService.resolveAttrs(this, tokenId, null) //Can a view have local attributes?
```
James-Sangalli, line 1092: An action can?

PR #1279, opened 2020-04-07
Fix tx to/from shown in transaction list and detail screens

JamesSmartCell:
Fix inconsistent display of sent and received in Transaction view.

Reported in the AlphaWallet public forum.

Issue can be seen by watching this wallet:

```0xB9E289bC3beAFd15397Ae3D03054864f1c519B74```

Be sure to enable Rinkeby. Before patch you'll see all tx are green Received, after patch they will show up correctly.


JamesSmartCell:
Thie PR fixes known bug reported on forum.

PR #1292, opened 2020-04-10
Refactor ERC721 token select. 

JamesSmartCell:
NOTE: This PR goes in AFTER 'Direct-DB-Load' #1291 

General code improvements - when reading from OpenSea we don't know what type of ERC721 so mark as ERC721_UNDETERMINED when we first see it. 
This type acts as a flag for the resolution code to know the type hasn't been fully identified yet.

This way of resolving makes the code cleaner.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/ContractType.java
+++ b/app/src/main/java/com/alphawallet/app/entity/ContractType.java
@@ -19,5 +19,6 @@ public enum ContractType
     DELETED_ACCOUNT,
     ERC721_LEGACY,
     ERC721_TICKET,
+    ERC721_UNDETERMINED, //when we receive an ERC721 we don't know what kind it is
     CREATION //Placeholder for generic, should be at end of list
 }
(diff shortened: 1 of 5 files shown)


James-Sangalli:
@JamesSmartCell when you say we don't know what "type" it is, are you referring to not knowing whether it is an ERC721 ticket or not? Please elaborate onto what this is trying to solve.

JamesSmartCell:
> when reading from OpenSea we don't know what type of ERC721

When we get the Token from OpenSea initially, we only know what we received is a type of ERC721.

We have 3 types of ERC721 - ERC721, ERC721_LEGACY and ERC721_TICKET

Since the TICKET variety is handled very differently (different class) to normal ERC721 we need to mark the type as UNDETERMINED until we have established exactly what type it is. This could be done inside the OpenSea fetch but then that class is performing activities outside of its scope, or we have to link in extra support. It's better to encapsulate that elsewhere.

It should have been handled this way to begin with really, but at the time I wanted to make the handling as small as possible.

JamesSmartCell:
To reproduce the bad behaviour:

Setup an emulator or make sure your AW install doesn't have your account starting with 0x42ea ...

Create a watch wallet for that address.

Look at the 'Star Wars' tickets at the bottom of the token list in the wallet page. It will switch between ERC20 and ERC721. Install this patch and it will very soon determine that the token is ERC721 Ticket and will stay set like this.

PR #1294, opened 2020-04-10
Fix database name case issue

JamesSmartCell:
NOTE: Goes in after #1293 Refator token type selection.

Ensure Realm files in the protected area have normalised names (all lower case).
When removing an account, delete all matching files (upper and lower case).

Can lead to multiple databases and false-negatives otherwise.

The wallet will self-correct once this PR goes through, and it resolves some update issues seen previously.


JamesSmartCell:
Can't use 'equalsIgnoreCase' because we're dealing with matching between a full path eg /0/data/data/io.stormbird.wallet/files/0x2A4bbC32(...)-db-.xml and an Address eg 0x2a4bbc32(...)

PR #1297, opened 2020-04-11
Improve ts card diagnostics

JamesSmartCell:
Report to TokenScript dev if an expected <view> isn't present in the script and display a token placeholder, rather than an empty space.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
@@ -10,6 +10,7 @@ import android.view.View;
 import android.widget.ProgressBar;
 import android.widget.TextView;
 
+import com.alphawallet.app.BuildConfig;
 import com.alphawallet.app.R;
 import com.alphawallet.app.entity.ContractType;
 import com.alphawallet.app.entity.TicketRangeElement;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
```diff
@@ -1109,12 +1128,41 @@ private void displayTicket(Context ctx, AssetDefinitionService assetService, Str
...
+     * @param range
+     * @param viewName
+     * @return
+     */
+    private String buildViewError(TicketRange range, String viewName)
     {
-        throwable.printStackTrace();
-        displayTicket(ctx, assetService, attrs, waitSpinner, tokenView, iconified);
+        String displayData = "<h3><span style=\"color:Green\">x" + range.tokenIds.size() + "</span><span style=\"color:Black\"> " + getTokenTitle() + "</span></h3>";
+        displayData += ("<br /><body>" + viewName + " fot found</body>");
```
James-Sangalli, line 1140: typo

James-Sangalli: Would be good to use localizable?

JamesSmartCell: good call, fixed both points.

PR #1307, opened 2020-04-13
Implement Gas Slider

### LINKED ISSUE #1271: Implement Gas Slider
Enhance the gas slider as a subunit to match the spec in Zeplin.

Spec is here:

https://zpl.io/b67kerd

However since this view isn't ready yet, the component can be written separately into the existing view 'activity_gas_settings' as a view component (similar to how the 'PasswordInputView' and 'FunctionButtonBar' work). 

This sub component can replace the old horrible looking one in GasSettingsActivity. It should look exactly like the component in the zeplin spec with 'Speed (Gas)' at the top and '1 Gwei      9000' at the bottom, and replaces the whole of the view in GasSettingsActivity, ie this:

separateview

This component should be ready to insert into the confirmation when this view is complete - so it'll need its own backing class same as FunctionButtonBar. Simply including the 'gas_selector' into your view you should get this self-contained component with auto-updating values. So, the class needs to regularly poll the API and update the view. It should also stop the polling when the view is destroyed.

You will need to convert Gwei to ETH for the central value (in the spec example 0.0003642). This obtained by multiplying the gas limit with the gas price as seen in the 'updateNetworkFee' calculation.

All values in the existing GasSettingsActivity are shown in Gwei (except network fee).

You'll need to know some basic conversion values, have a look at this site and try a few values: 

https://converter.murkin.me/

class BalanceUtils has some basic conversions written in code you can use.

All calculations should be done in wei and then scaled back to ETH.

Example: I set the slider gas price at 1.2 Gwei (1.2 * 10^9 wei) and gas limit to 90000 Gwei (90000 * 10^9 wei). Expected network fee is 90000 Gwei or 90000000000000 wei which is 0.00009 ETH.

Initially, to get implementation working quick set a default gas limit to 90000 Gwei. You can later update this with a method but let's get it working first.

Why do we want to change gas and what is it? Gas is the network transaction fee. It's an incentive for Ethereum miners to include your transaction in their block calculation. If they win the mining competition in addition to mining reward they get your gas fee if they included it in their block. 

If there's a lot of transactions competing for limited block space the miner will always want to include the transaction with highest included gas fee.

Therefore there's an expected transaction time associated with the fee, the higher the fee you include the more quickly your transaction gets processed. There's an API you can call with gas fee vs expected inclusion time here: 

https://ethgasstation.info/json/ethgasAPI.json

All gas prices are in 10*Gwei, so you'll see fast = 50, safeLow = 10 which is 5 Gwei and 1 Gwei.

```
  "fast": 50.0,
    "fastest": 100.0,
    "safeLow": 10.0,
    "average": 10.0,
```
...
```
corresponding speed in minutes

    "safeLowWait": 2.8,
    "avgWait": 2.8,
    "fastWait": 0.3,
    "fastestWait": 0.3,
```

These are prices in gwei * 10, so eg for safeLow, gas price is 1 gwei, expected wait time is 2 minutes 48 seconds (2.8 mins).

Let's say the gas limit is set at 90000 Gwei, and I want my transaction to be fast I set the slider at 5 Gwei. The Expected fee will be 90000*5/10^9 = 0.00045 ETH. My expected wait time for this is 0.3 mins or 18 seconds.

Make sense? Lets use safeLow as the minimum slider position and 'fastest' as the maximum slider position. You can use linear interpolation to calculate the estimated transaction time for now. It's more complex than linear but let's enhance that later. 

Addend: Once the slider is working with linear interpolation the enhancement is to use the more detailed information from the API to produce more accurate wait time:

```
    "gasPriceRange": {
        "100": 0.3,
        "95": 0.3,
        "90": 0.3,
        "85": 0.3,
        "80": 0.3,
        "75": 0.3,
        "70": 0.3,
        "65": 0.3,
        "60": 0.3,
        "55": 0.3,
        "50": 0.3,
        "45": 0.7,
        "40": 0.7,
        "35": 0.7,
        "30": 0.8,
        "25": 1.0,
        "20": 1.0,
        "15": 1.4,
        "10": 2.8,
        "8": 133.9,
        "6": 133.9,
        "4": 133.9
    }
```

Once you finished getting the slider working you will understand the above values - they are just gas price vs expected wait time. From bottom, 0.4 Gwei = 133 minutes 54 seconds, 10 Gwei = 18 seconds. Use linear interpolation between those finer point values, you will end up with a time spread, which if represented visually will be like the view in MetaMask:

gascap1

You can see the enhancement they make too: since above 5 Gwei there's no improvement they don't allow you to slide above 5. This is definitely an enhancement we want to make too.

ChintanRathod:
Fixes #1271


ChintanRathod:
@JamesSmartCell Component is working as expected. I just replaced GasSetting screen with newly created component but didn't yet integrate with confirmation screen. 
Shall I replace existing Gas Price and Gas Limit logic and align with this component?

JamesSmartCell:
Yes, please hook up the existing methods on creation.

Also can you refactor the layouts with @justindg's new dynamic sizing attributes -

eg ```android:textSize="17sp"``` becomes ```android:textSize="@dimen/sp17"```

```android:layout_marginEnd="25dp"``` becomes ```android:layout_marginEnd="@dimen/dp25"```

If the required param doesn't exist yet, add it in and make it the same across all the different dimensions (or try to scale it proportional to what's already in the files); we can re-size the margins/font sizes for the different screen sizes later during internal testing.

This is to auto-resize components for different sized Android display screens.

And, when adding strings resources please tick the localisation boxes for 'es' and 'zh':

image

If you have already added the string, then simply copy it into the relevant directory. This helps with localisation, we can just send the strings files off to our translators (or translate in-house).

ChintanRathod:
- [x ] String resources in other translation files 
- [ x] dimension update for multiple resolution
@JamesSmartCell Added code for above points. Pending with old layout removal GasSettingActivity.

JamesSmartCell:
I updated the code with a few points:

- Remove replaced views
- Replace button with standard button bar (so we can change all the buttons in one go if the design changes)
- Hook up the gas price and gas limit from slider view to the 'save' function
- min/max the values of received data for improved user experience
- extrapolate between received values

If you work on a view; please ensure you use the FunctionButtonBar to place function buttons at the bottom if it doesn't already.

PR #1344, opened 2020-04-20
Token script management

### LINKED ISSUE #1315: Add TokenScript Management Page
Page that matches iOS TokenScript management page which displays a list of Tokens that have a tokenscript.

Once we have a basic list of tokens we will expand on it to show warnings/errors and allow users to remove and otherwise manage scripts.

Add the page into the settings like this (@hboon can you please add iOS settings page screen shot with TokenScript management entry and another showing the script page):

TokenScript Management

You will be able to get a list of the scripts from assetDefinitionService using a method that traverses the ```assetDefinitions``` SparseArray/Mapping, making a list of holding contract tokens vs Filename. See method 'addContractAddresses' to see how to extract the holding contracts from each definition. Most scripts have only one holding contract token (eg Karma - ```0x75cb...```).

I'm thinking initially we want to see something like a list like this:

```
Script Name                 Filename                        
   - token address
```
eg
```
Car Ownership Token                 karma.canonicalized.xml (small font)
0x75cb...  (small font)
```

We may want to change this in future but for now this will be fine

ChintanRathod:
Fixes #1315


ChintanRathod:
@JamesSmartCell Added code to display the information about the TokenScript files. There are total 4 files visible on the screen like below. 

image

Path : /xdaicanonicalized.xml
Path : /data/user/0/io.stormbird.wallet/files/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48.xml
Path : /data/user/0/io.stormbird.wallet/files/0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359.xml
Path : /storage/emulated/0/AlphaWallet/karma.canonicalized.xml

But I applied logic to only display items from "AlphaWallet" directory and below will be output after filtering entries. 

image

- [ ] Confirmation on above implementation
- [ ] Need image icon to set at "TokenScript Management"

JamesSmartCell:
@ChintanRathod please revert changes to AssetDefinitionService, then we can get this into the source.

JamesSmartCell:
@ChintanRathod bump.

ChintanRathod:
@JamesSmartCell Could you please help me to understand what to revert from the AssetDefinitionService, because if I revert the changes, this branch won't have fruitful output.

JamesSmartCell:
I mean to revert the changes to the file 'AssetDefinitionSevice.java'. It appears in your commit that there are only formatting changes, but it's difficult to be sure. Can you revert the formatting changes so it's clear whether there's any change to this specific file.

PR #1385, opened 2020-05-07
changing default accent colour following fiji design

colourful-land:
this shouldn't have much visible effect since barely any code use accentColor, but old accent colour should be replaced.


colourful-land:
the PR is too small, retracting to make another.

PR #1386, opened 2020-05-08
Update with constraint layout

### LINKED ISSUE #1351: Use adaptive layout for 'Redeem' activity.
Currently Redeem layout crushes the TokenScript view on certain smaller sized phone displays.

Convert the view to an adaptive one so that smaller phones still display correctly. See attached picture from Pixel 1st gen. Maybe use ConstraintLayout to build a relative view.

The relevant activity is RedeemSignatureDisplayActivity.

1a39c1ce-4504-4dc2-af3c-e2aa9decdb75

ChintanRathod:
Fixes #1351

### DIFF
--- a/app/src/main/res/layout/activity_rotating_signature.xml
+++ b/app/src/main/res/layout/activity_rotating_signature.xml
@@ -1,92 +1,88 @@
 <?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<android.support.constraint.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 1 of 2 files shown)


ChintanRathod:
Output is like below. Kindly review and revert if it is as expected. I didn't work on QR code resize as it won't make too much difference except Top padding and Bottom padding. 

scrolling screen

JamesSmartCell:
Is it possible to have the view re-size the token instead? Scroll is not a good solution in this case.

JamesSmartCell:
OK it may not be possible to scale the webview. Try this: If the screen is detected to be smaller then don't display the 'present the to redeem ...' (can't remember the exact phrase, but the text that appears at the top of the screen below the title), only display the QR code at the top, then the token at the bottom of the screen - could use RelativeLayout, although ConstraintLayout would be better.

ChintanRathod:
@JamesSmartCell Replaced existing view with Constraint layout as suggested. Output is like below in Pixel 1. 

image

PR #1397, opened 2020-05-13
Implement local-ref tags

JamesSmartCell:
Implement the local-ref tags part of the schema 2020-03.

This PR adds handling of local tags and also does a minor refactor of the parser and TS function handling to improve code and handling.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
@@ -76,6 +76,8 @@ public class Token implements Parcelable, Comparable<Token>
     public int nonIconifiedWebviewHeight;
     private int nameWeight;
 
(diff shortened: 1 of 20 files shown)


James-Sangalli:
> @JamesSmartCell two issues:
> 
> 1. The email record for bitcoinwarrior1.eth no longer resolves
> 2. the renew button triggers the following error:
> 
> > I/chromium: [INFO:CONSOLE(1)] "Uncaught ReferenceError: onConfirm is not defined", source: data:text/html; charset=utf-8;base64,PGhlYWQ+PG1ldGEgbmFtZT0idmlld3BvcnQiIGNvbnRlbnQ9IndpZHRoPWRldmljZS13aWR0aCwg%0AaW5pdGlhbC1zY2FsZT0xLCBtY....

Edit: number one was due to the TS, once fixed it resolves but it shows the old record rather than the new one: https://etherscan.io/tx/0xe76d1c55776ada393875765a1f41a60f70b9294812c579b6b2c3ae13a1abde39

The renew button is still stuck with the error above.

PR #1431, opened 2020-05-27
Api23 Final

JamesSmartCell:
This branch is the code for the final API23 release. Only significant bug fixes will be applied.


Shutch748 (changes requested):
Not yours

JamesSmartCell:
Now 100% irrelevant

PR #1447, opened 2020-06-03
Enhance TokenScript Management panel

JamesSmartCell:
Expand TokenScript management panel functionality; 

- show overridden scripts.
- show all origin contracts within each script and in-detail script override (on click).
- allow user to delete scripts (on long click).


JamesSmartCell:
You can delete server files, but they will come back. This could be a good support tool in future to be absolutely sure a user has the latest script, like 'refreshing TokenScript cache'. Now that the script load is more closely coupled with viewing tokens it makes the loading of scripts more predictable.

PR #1459, opened 2020-06-15
Unstoppabledomains

JohnnyJumper:
This allows all alpha-wallet users to resolve .crypto domains. 

The library is making an http request to a blockchain provider (I chose infura as the most popular blockchain provider on the market now and created an api key for such. Alpha wallet might be interested in updating the key to their own which should be quite easy) Another option would be to connect your own node to the library. (just provide it with url to make a post request in json-rpc format and it should work) 

I couldn't understand if there are any other currencies that are supported beside eth (like btc, lite, dodge and etc).
In case if there is the support is very intuitive and easy as we need to pass the ticker name to this line `this.resolution.addr(domain[0], "eth");`

Also, this is might be not the ideal but it is the first step and I am willing to discuss the next iterations.


JamesSmartCell:
Hi @JohnnyJumper thanks for this addition we'd like to get this into the source. Could you provide an example .crypto name that resolves into an ethereum address? I wasn't able to find any that worked, but not sure which services support .crypto.

JohnnyJumper:
Hi @JamesSmartCell  
Try to resolve brad.crypto. It should resolve to '0x8aaD44321A86b170879d7A244c1e8d360c99DdA8',

JamesSmartCell:
Works, great. I'm doing a small refactor in my code before pushing as it's a bit messy. Can you link to the contract call details so we can implement it on the iOS build of our app? Or, if you or someone on your team know Swift can you submit a PR to the iOS repo?

JamesSmartCell:
Can we get .crypto domains from a contract call? If so, we'd like to make a TokenScript to help people obtain them. Or, do you have it set up so only admin can call those functions, and you use fiat to buy the domain?

JamesSmartCell:
One more thing - how do we do a reverse resolve? That is, if Brad, the owner of brad.crypto imports his wallet into AlphaWallet, can we see if there's a .crypto associated with that address and if so, name the wallet appropriately?

JohnnyJumper:
Hi, Sorry for the messiness of the code, Android development and swift is something I never touched before as you deducted above. 

Regarding the contract call we were using the [resolution library](https://github.com/unstoppabledomains/resolution-java) to make an http call to a blockchain provider calling the following methods and contracts in order:

1. Call the registry contract to get the resolver one
Registry contract address: 0xD1E5b0FF1287aA9f9A268759062E4Ab08b9Dacbe
method: resolverOf
Arguments: namehash of the domain (same as in ens)
This should return you the address for resolver contract

2. Call the resolver contract to get the address for the currency needed:

method: get
arguments: [ key, namehash ]
String key = "crypto." + ticker.toUpperCase() + ".address";   
the ticker is "eth" or "btc" or any other coin ticker.

In case if returned nulled or errored out we also make a call to registry contract ownerOf method with argument namehash of a domain to get the owner address -> if that is missing then we throw the error saying that such domain is not registered, otherwise we throwing an error saying that such coin is not attached to the domain

I found some tool to convert [Java library]( https://github.com/google/j2objc )  to objective C
Also looks like the Java to Swift is in development [link](https://github.com/SwiftJava/SwiftJava/blob/master/README.md) 
maybe they can help to speed up the process? 

Regarding the reverse resolution currently, we only support it on API level for domain resellers see [docs](https://docs.unstoppabledomains.com/#tag/reseller/paths/~1resellers~1{resellerID}~1domains/get)
There are some security issues with reverse domains without verification.
Imagine someone set up an eth address for someone else on a domain since domain usage is easier than pure eth addresses. Later the owner of the domain can just go and change the addresses without knowledge. Not everyone fully understands this and some people might be using it as an advantage. We do have a plan in long term to provide the domain verification and after that one, we can start focusing on implementing the reverse lookup on the contract level.

JamesSmartCell:
The messiness was in my old code - yours is fine :) When I looked at it I needed to refator.

I would like to update your PR before I pull. Can I either send to diff to you or can you add me as a collaborator on your fork?

JamesSmartCell:
> 1. Call the registry contract to get the resolver one
>    Registry contract address: 0xD1E5b0FF1287aA9f9A268759062E4Ab08b9Dacbe
>    method: resolverOf
>    Arguments: namehash of the domain (same as in ens)
>    This should return you the address for resolver contract
> 2. Call the resolver contract to get the address for the currency needed:
> 
> method: get
> arguments: [ key, namehash ]
> String key = "crypto." + ticker.toUpperCase() + ".address";
> the ticker is "eth" or "btc" or any other coin ticker.

thanks that will help

JohnnyJumper:
I have added you as a collaborator for the fork as it seems the easiest solution. If there is something I can do to help please make sure to tell me. We all are interested in seeing alpha wallet as one of our resolution partners

JamesSmartCell:
Thanks, I'll add the contract call details to an issue for the iOS team.

JohnnyJumper:
Thank you so much for the fast response!

JamesSmartCell:
Hi @JohnnyJumper , I noticed the unstoppable library doesn't seem to work with Android 9; there's a Java type that's undefined - it works fine with Android 10. I called the contracts manually using web3j as you described above, which works fine.

JohnnyJumper:
Oh that's good to know! I will look into it

JohnnyJumper:
@JamesSmartCell Can you point me on how to check for error so I can fix it?

PR #1462, opened 2020-06-15
Keys picked up from project root

JamesSmartCell:
Allow keys file to go into project root directory, this helps with multiple projects.


JamesSmartCell:
Only pick up keys from project root directory @colourful-land

PR #1467, opened 2020-06-16
TokenScript error show in dialog

### LINKED ISSUE #1449: Show errors for each tokenscript
When the TokenScripts are loaded for the Management screen, any errors are thrown as ParseExceptions and are caught and currently nothing is done with them:

```
public Single<List<TokenLocator>> getAllTokenDefinitions(boolean refresh)
...

                        } // TODO: Catch specific tokenscript parse errors to report tokenscript errors.
                        catch (Exception e)
                        {
                            // don't add to list
                        }
```
These should be added to the list, but labelled as error (replace the 'lock' image in the corner with the error asset - a red circle with the cross in it, this asset is already in the resources somewhere).

The error that was reported should be shown when you click on the TokenScript in the pop-up dialog (the one that shows when you click on the TokenScript - normally it shows the origin contracts for each TokenScript).

Note that due to the way an exception is thrown and parsing stops at the error; we will only show one error - debugging initially will be an iterative process.

This will be a huge help debugging TokenScripts.

There's no design for this, so feel free to just show the error text in the popup window; it's for developers rather than end users so it just needs to be functional.

ChintanRathod:
Fixes #1449

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/TokenLocator.java
+++ b/app/src/main/java/com/alphawallet/app/entity/TokenLocator.java
@@ -7,11 +7,23 @@ public class TokenLocator extends ContractInfo
 {
     private final TokenScriptFile tokenScriptFile;
     private final String name;
+    private boolean error;
+    private final String errorMessage;
 
     public TokenLocator(String name, ContractInfo origins, TokenScriptFile file) {
         super(origins.contractInterface, origins.addresses);
         this.name = name;
         this.tokenScriptFile = file;
+        this.error = false;
+        this.errorMessage = "";
+    }
+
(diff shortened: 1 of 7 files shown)


ChintanRathod:
Output of current implementation

ezgif com-video-to-gif (1)

JamesSmartCell:
Almost works 100%. Just a couple of issues I found:

1. When there's an error, the token name doesn't get blanked, so RecyclerView leaves the token name from the previous view:

TSerror1



2. When I click on the error, the leading text doesn't make sense within the context (TokenScript overriden by .....), use something like 'TokenScript Error: ':



TSerror

ChintanRathod:
@JamesSmartCell Feedback applied. Could you please check.

PR #1469, opened 2020-06-17
Check existing ENS name stored in preference

### LINKED ISSUE #1404: Check existing ENS sends when using reverse resolve
When we reverse resolve ENS (ie look up an address to see if it has an ENS name associated) using:
```resolveEnsName(String address)```, if we don't get any result from the ```reverseResolve``` method we should also check the previously used ENS names which are retrieved here: ```getENSHistoryFromPrefs();``` in ```class AutoCompleteUrlAdapter```.

If we resolve the address for those (using ```resolveENSAddress(String name)```, and it matches the address we're looking up then we found a hit. This way a user can discover their own wallet names for the wallets list, if they haven't (or don't know how to) set the reverse resolver contract.

```class AutoCompleteUrlAdapter``` should probably also be renamed to ```class AutoCompleteAddressAdapter```, it no longer has anything to do with URLs.

ChintanRathod:
Fixes #1404

### DIFF
--- a/app/src/main/java/com/alphawallet/app/C.java
+++ b/app/src/main/java/com/alphawallet/app/C.java
@@ -169,6 +169,7 @@ public abstract class C {
 
     public static final String ENS_SCAN_BLOCK = "ens_check_block";
     public static final String ENS_HISTORY = "ensHistory";
+    public static final String ENS_HISTORY_PAIR = "ens_history_pair";
 
     public enum TokenStatus {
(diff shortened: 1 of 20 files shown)


ChintanRathod:
I changed a flow bit, rather checking after we fail to resolve ENS, we check for already stored ENS name, in this way if app has already reverse resolve the ENS name before, no need to call network for another time.

PR #1472, opened 2020-06-21
Add speedup and cancel transaction feature

### LINKED ISSUE #1430: Gas price +1
Can we be a bit more aggressive with the gas price? If the node returns with a gas price n, then we should use n+1 as the gas price.

JamesSmartCell:
While transaction is still pending, allow the feature to 'add gas' or cancel the transaction.

Also:
- show the gas price used for transactions in the detail view.
- live update how long the transaction has been pending for.
- live update how many confirmations the transaction has.

Closes #1430


JamesSmartCell:
@colourful-land can you re-pull and check the issues are fixed.

colourful-land:
> @colourful-land can you re-pull and check the issues are fixed.

Still buggy for me.

1. Send a TX. Give it little gas.
2. Cancel it.
3. Speed it up (speed up the cancellation)
4. You get errer that's already written to the blockchain when it isn't

(I waited 30 mins it still isn't written to the blockchain by observing the dancing circle in TX history and the UI still lies to me)

record

JamesSmartCell:
Is the problem repeatable? I wasn't able to reproduce. Could you outline a repeatable way to create this problem? Does it happen if you try again?

colourful-land:
> Is the problem repeatable? I wasn't able to reproduce. Could you outline a repeatable way to create this problem? Does it happen if you try again?

I already did:



    Send a TX. Give it little gas.
    Cancel it.
    Speed it up (speed up the cancellation)
    You get errer that's already written to the blockchain when it isn't

and I repeated it again and successfully created the errorr again. It's like one shot one kill - you can't fail to produce the error

colourful-land:
> Is the problem repeatable? I wasn't able to reproduce. Could you outline a repeatable way to create this problem? Does it happen if you try again?

Maybe you failed to notice that it's about speeding up a cancellation?

JamesSmartCell:
Thanks got it!

colourful-land:
> Thanks got it!

If you are afraid of more merge conflict later,
 you can rebase this

PR #1478, opened 2020-06-23
only check events for tokens with non-zero/non-empty balance

JamesSmartCell:
Don't bother checking events for tokens with zero balance.


colourful-land:
Are you sure:

1. this only affects none-fungible tokens
2. that this change does not affect the case where the user has 0 crypto-kitty and he got sent 1?

JamesSmartCell:
> Are you sure:
> 
> 1. this only affects none-fungible tokens
> 2. that this change does not affect the case where the user has 0 crypto-kitty and he got sent 1?

1. It now only affects NFTs.
2. This particular change is picked up from OpenSea query not from events. We would only pick up the balance change event if the script had added it; which is an impossible assumption as you would need to replicate it for every ERC721. Far simpler to rely on the author registering with OpenSea.

PR #1482, opened 2020-06-25
Prevent keyboard hiding webpage

JamesSmartCell:
Fix issue where soft keyboard covers webpage areas.

Fix related issues:

- clear up old issue with bottom of webpage last transaction being hidden by the bottom nav bar (issue since 'sliding view' (in wallet view that covers the title when you slide up) was introduced).
- Refactor the keyboard presence sensing in PasswordInputView using the keyboard event library. This removes some of the onscreen clutter (buttons, info) when the keyboard is active - needed on lower resolution devices.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -140,6 +140,9 @@ dependencies {
     implementation "io.reactivex.rxjava2:rxjava:$project.rxJavaVersion"
     implementation "io.reactivex.rxjava2:rxandroid:$project.rxAndroidVersion"
 
+    // Keyboard visibility
+    implementation 'net.yslibrary.keyboardvisibilityevent:keyboardvisibilityevent:3.0.0-RC2'
+
     // Dagger 2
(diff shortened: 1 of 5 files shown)


James-Sangalli:
@JamesSmartCell how to reproduce this problem and what is the problem exactly?

JamesSmartCell:
Find a website where you need to enter information, good example is: https://shop.qoin.world/register/

Now click on the username to get the soft keyboard appearing. Notice how the keyboard covers the bottom part of the website, so if there were further inputs or information you can't see them until you clear the input.

Now try with this code. See how it now scrolls correctly. Also, there was a small issue with the bottom of the website getting cut off since the sliding wallet view was introduced in Fiji. You can best see this on the POA Mania dapp - https://www.poamania.com/ See the bottom of the page where it says 'Join us:', you can't see the text of how to join.

PR #1483, opened 2020-06-25
Add universal scanning

### LINKED ISSUE #1388: Add 'Universal Scan' button
On the wallet front page there should be a scan button next to 'add token' which looks like this [+] (see zeplin for asset - use SVG asset - don't generate many PNG's).

The result should be passed through the QRUrlDecoder and then the app assigns where to go based on the result.

There should only be three actions - if the EIP681Type is URL then open it in Dapp browser, if OTHER then fail, if anything else then jump to 'SendActivity' and handle appropriately.

ChintanRathod:
Fixes #1388


ChintanRathod:
@JamesSmartCell This PR is missing with Camera buttons on the bottom i.e. Light, My Address and Browse. 
Shall I include it in this PR only or it should cover in separate issue.

JamesSmartCell:
Hi @ChintanRathod this looks good. Have you replaced the QR scanning screen with a new activity? If so could you remove the old one and use your new one in all the places where QR scanning is done for consistency.

I will move the code to detect a universal/magic link into the parser, to make the code a bit cleaner.

We can add the other features to a separate PR.

ChintanRathod:
@JamesSmartCell I actually use old activity with a checkmark to use as a Universal Scanner. So for now, we don't need to remove it. I also check for different option to integrate new library but doesn't seem very good with new system support so I kept old logic only. 

image

JamesSmartCell:
Also can you add the new "crypto" domain to the 'couldBeENS' method in the QRURLParser

JamesSmartCell:
One of the tests are failing- probably just a case of adding the new virtual method into the test.

ChintanRathod:
> One of the tests are failing- probably just a case of adding the new virtual method into the test.

Test cases are passed.
Changes made as you suggested for HomeActivity. I kept logic in ViewModel.

PR #1484, opened 2020-06-25
add type to asnx:module syntax

James-Sangalli:
AlphaWallet/TokenScript-Examples#93


colourful-land:
@JamesSmartCell can you check to be sure that it this should work for the cases both with type and without? That is, before and after the application of https://github.com/AlphaWallet/TokenScript-Examples/pull/93/files

JamesSmartCell:
Now works with both schema types.

PR #1489, opened 2020-06-29
Flow changed for "+" icons on home screen

### LINKED ISSUE #1485: Get rid of two "+" buttons from the wallet tab
@ChintanRathod 

We need a quick change. 
Currently, we have two "plus" icons in the Wallet tab. This is very confusing, but we can fix this quite easy and quickly. 

So we need to change what follows:

1. Take out the very top + button and replace it by QR icon https://zpl.io/VYvX6Ge
2. This QR code icon will open "My Wallet Address". We already have this screen in Settings tab. So now, we will have two ways to access the same screen. First is QR Code icon in the Wallet tab, second is "My wallet Address" item in the Settings tab.
3. Currently, to add a Custom Token you can click very top "+" button. Now, you can open "Add /Hide Tokens" and there you will find + icon. 
4. Change text from "Show / Hide Tokens" to "Add / Hide Tokens"

Plus

ChintanRathod:
Fixes #1485

### DIFF
--- a/app/src/main/java/com/alphawallet/app/di/TokenManagementModule.java
+++ b/app/src/main/java/com/alphawallet/app/di/TokenManagementModule.java
@@ -2,6 +2,7 @@ package com.alphawallet.app.di;
 
 import com.alphawallet.app.interact.ChangeTokenEnableInteract;
 import com.alphawallet.app.repository.TokenRepositoryType;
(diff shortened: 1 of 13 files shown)


ChintanRathod:
@tomekalphawallet Below is output of current implementation. 

ezgif com-video-to-gif

PR #1492, opened 2020-06-30
Add Firebase analytics

### LINKED ISSUE #1476: Analytics for AlphaWallet
Apply Firebase analytics to base AlphaWallet.

Duplicate the work done for the Qoin fork on the main repo.

You will need to liase with @hboon to ensure only the features agreed upon for the main branch are implemented.

Specifically:

- Events need to be anonymous - no addresses, transaction hashes etc.

ChintanRathod:
Fixes #1476


ChintanRathod:
@JamesSmartCell 

1) I created another flavor which is "noAnalytics" and it has conditional gradle dependency integration which will not integrate Google Plugin and Firebase analytics library. 

2) Please have a note that I downgraded Firebase analytics dependency version to "16.5.0" where recent one is around "17.4.3" which is only compatible with "AndroidX" support library. 

Ref: https://firebase.google.com/support/release-notes/android#2019-06-17

I was facing duplicate class error and found solution hence for now we need to deal with lower version. Once we have AndroidX integration we will update it.

JamesSmartCell:
Can we tidy the flavour up a bit?

How about a single file that's different between the flavours. Abstract the calls into that file via an interface. If Analytics is in the build then the file has the functions filled in, in the non flavoured build the functions are empty. This keeps the source base a lot cleaner. Currently there's too much code duplication for this to be a valid solution - leading to many maintenance issues further down the line.

I also ran into the AndroidX/Firebase issue previously. We can't upgrade that until we do the AndroidX upgrade. I have a branch that's been AndroidX'd but needs more testing.

Also, Travis build check is failing on this checkin.

ChintanRathod:
@JamesSmartCell 
I understand your concern but if we create a flavor, 
1) We have to put duplicate files which are in "awallet" flavor i.e. 
image
2) Point here is integrating "google" plugin and "firebase analytics library" dependency. We might run in to "Class not found" error if we do conditional integration and we are bifurcating in single class. 

If we don't have an issue to put "gradle plugin" and "firebase analytics" to module gradle file without condition, we can manage to bifurcate with some BOOLEAN check mark in gradle file without any "flavor"

ChintanRathod:
About the Travis job fail, I cross check the issue it is facing and below is output. 
image
The solution is to modify Travis configuration file to overcome this.
https://talk.openmrs.org/t/core-apps-travis-build-failure-the-job-exceeded-the-maximum-log-length-and-has-been-terminated/23774/39

JamesSmartCell:
Yes, understood about the flavour. Let's not do it this way then, it's creating too much extra baggage that will be a maintenance issue in the future.
Can you ensure all the code that talks to Firebase is in one file, which we can simply template in and out - so, the file would look something like this:
Not ideal, but it looks like anything else is going to be an albatross around our necks.

class Analyse
{
    public void handleSomeEvent() { call to firebase }
} 

/*
class Analyse
{
    public void handleSomeEvent() {  }
}
*/

Within each class, have an assert that detects the build flag in gradle and fires if it's wrong - that is, if the build is with firebase, and the class which does not call firebase is uncommented, it should fire and vice-versa.

JamesSmartCell:
> About the Travis job fail, I cross check the issue it is facing and below is output.
> image
> The solution is to modify Travis configuration file to overcome this.
> https://talk.openmrs.org/t/core-apps-travis-build-failure-the-job-exceeded-the-maximum-log-length-and-has-been-terminated/23774/39

Why does it appear in this branch? Is is because of the additional flavour?

ChintanRathod:
> Why does it appear in this branch? Is is because of the additional flavour?

I guess due to google play service plugin and firebase dependency.

ChintanRathod:
@JamesSmartCell Feedback changes applied, could you please check

- Global plugin and library implementation
- DI based on the BuildConfig parameter "USE_ANALYTICS"

PR #1494, opened 2020-07-02
Implement Gas Estimation

### LINKED ISSUE #1471: Use Web3j Estimate gas to get a more accurate gas estimate
I wrote an outline method to estimate gas for transaction calls (it doesn't work because of network on main thread and also needs to be broken up via flow):

```
    //Outline only - will need to be broken up via flow
    public void getGasEstimate(byte[] transactionBytes, int chainId, String to, BigInteger amount)
    {
        String txData = "";
        if (transactionBytes != null && transactionBytes.length > 0)
        {
            txData = Numeric.toHexString(transactionBytes);
        }

        Web3j web3j = TokenRepository.getWeb3jService(chainId);

        BigInteger nonce = ethereumNetworkRepository.getLastTransactionNonce(web3j, defaultWallet.getValue().address); //latest nonce for wallet & chainId

        Transaction thisTx = new Transaction(
                defaultWallet.getValue().address,
                nonce,
                gasService.getGasPrice(),
                gasService.getGasLimit(),
                to,
                amount,
                txData);

        
        EthEstimateGas gas = web3j.ethEstimateGas(thisTx).send();

        //gas estimate (BigInteger):
        if (gas.getError() == null)
        {
              gasEstimate = gas.getAmountUsed();
        }
    }
```

This call would go in class ConfirmationViewModel, and replaces the estimate heuristic in GasService:

```
    private BigInteger estimateGasLimit(byte[] data)
    {
        BigInteger roundingFactor = BigInteger.valueOf(10000);
        BigInteger txMin = BigInteger.valueOf(C.GAS_LIMIT_MIN).multiply(BigInteger.valueOf(5));
        BigInteger bytePrice = BigInteger.valueOf(C.GAS_PER_BYTE);
        BigInteger dataLength = BigInteger.valueOf(data.length);
        BigInteger estimate = bytePrice.multiply(dataLength).add(txMin);
        estimate = estimate.divide(roundingFactor).add(BigInteger.ONE).multiply(roundingFactor);
        return estimate;
    }
```

Which is a 'naïve' implementation that doesn't take account of loops, special functions etc.

You will need to check 
```
if (gas.getError() == null) 
{
       //got a good estimate
      BigInteger gasEstimate = gas.getAmountUsed();
}
else
{
      //Transaction looks like it will fail - notify user; pass the String error message we receive here to user
}
```

You'll need to add the EthereumNetworkRepository service to the viewmodel, as well as pre-building the transaction parameters properly (done in finaliseTransaction() ) after the view is created.

ChintanRathod:
Fixes #1471

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokens/Token.java
@@ -176,7 +176,7 @@ public class Token implements Parcelable, Comparable<Token>
         {
             for (BigInteger id : getArrayBalance())
             {
-                if (!uniqueIds.contains(id)) uniqueIds.add(id);
+                if (!uniqueIds.contains(id) && !id.equals(BigInteger.ZERO)) uniqueIds.add(id);
             }
         }
         else
--- a/app/src/main/java/com/alphawallet/app/entity/tokenscript/EventUtils.java
+++ b/app/src/main/java/com/alphawallet/app/entity/tokenscript/EventUtils.java
@@ -620,7 +620,7 @@ public abstract class EventUtils
     private Event generateEventFunction(EventDefinition ev)
     {
         List<TypeReference<?>> eventArgSpec = EventUtils.generateFunctionDefinition(ev.eventModule.getSequenceArgs());
-        return new Event(ev.eventName, eventArgSpec);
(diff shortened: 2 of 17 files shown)


ChintanRathod:
Output 

ezgif com-video-to-gif

PR #1499, opened 2020-07-07
Add update revamp

### LINKED ISSUE #1487: Update "Add / Hide Tokens" screen
Zeplin:

This is how it looks now:

---

ChintanRathod:
Fixes #1487

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/TokenManagementActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/TokenManagementActivity.java
@@ -3,20 +3,24 @@ package com.alphawallet.app.ui;
 import android.arch.lifecycle.ViewModelProviders;
 import android.content.SharedPreferences;
 import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
 import android.preference.PreferenceManager;
 import android.support.annotation.Nullable;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
+import android.text.Editable;
+import android.text.TextWatcher;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.widget.Button;
 import android.widget.CheckBox;
+import android.widget.EditText;
 
 import com.alphawallet.app.C;
 import com.alphawallet.app.R;
 import com.alphawallet.app.entity.Wallet;
 import com.alphawallet.app.entity.tokens.Token;
-import com.alphawallet.app.repository.EthereumNetworkRepository;
 import com.alphawallet.app.router.HomeRouter;
 import com.alphawallet.app.ui.widget.adapter.TokenListAdapter;
 import com.alphawallet.app.viewmodel.TokenManagementViewModel;
@@ -27,17 +31,21 @@ import javax.inject.Inject;
 import dagger.android.AndroidInjection;
 
 public class TokenManagementActivity extends BaseActivity implements TokenListAdapter.ItemClickListener {
+
     @Inject
     TokenManagementViewModelFactory viewModelFactory;
-
     private TokenManagementViewModel viewModel;
 
     private RecyclerView tokenList;
     private Button saveButton;
-    private TokenListAdapter adapter;
+    private EditText search;
     private CheckBox hideZeroBalanceCheckBox;
+    private TokenListAdapter adapter;
 
     private Wallet wallet;
+
+    private Handler delayHandler = new Handler(Looper.getMainLooper());
+    private Runnable workRunnable;
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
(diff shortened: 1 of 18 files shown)


ChintanRathod:
Output

ezgif com-video-to-gif (2)

JamesSmartCell:
What does the 'Jake Wharton' rx library do in this case? is it associated with the search feature?

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokenListAdapter.java
```diff
@@ -1,113 +1,339 @@
...
+public class TokenListAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> implements Filterable {
+    private LayoutInflater inflater;
+    private Context context;
+    private ItemClickListener listener;
+    private final List<Token> tokens;
+    private ArrayList<Object> filteredData = new ArrayList<>();
+    private ArrayList<TokenSortedItem> displayTokensList = new ArrayList<>();
+    private ArrayList<TokenSortedItem> hiddenTokensList = new ArrayList<>();
+    protected final AssetDefinitionService assetService;
+
```
JamesSmartCell, line 52: Deduplicate.

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokenListAdapter.java
```diff
@@ -1,113 +1,339 @@
...
+                    }else{
+                        hiddenTokensList.add(new TokenSortedItem(TokenViewHolder.VIEW_TYPE, t, t.getNameWeight()));
+                    }
                 }
             }
         }
 
-        Collections.sort(filteredList);
-        return filteredList;
+        if(performSorting) {
```
JamesSmartCell, line 81: inconsistent formatting

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokenListAdapter.java
```diff
@@ -1,113 +1,339 @@
...
+        {
+            TokenSortedItem token = (TokenSortedItem) filteredData.get(position);
+            TokenViewHolder tokenViewHolder = (TokenViewHolder) viewHolder;
+            viewHolder.setIsRecyclable(false);
+            tokenViewHolder.tokenName.setText(token.value.getFullName());
+            tokenViewHolder.switchEnabled.setChecked(token.value.tokenInfo.isEnabled);
+            tokenViewHolder.switchEnabled.setTag(new Integer(position));
+            tokenViewHolder.switchEnabled.setOnCheckedChangeListener((v, b) -> {
+                listener.onItemClick(((TokenSortedItem) filteredData.get(position)).value, b);
+
```
JamesSmartCell, line 153: formatting

app/src/main/java/com/alphawallet/app/ui/widget/holder/TokenHolder.java
```diff
@@ -260,24 +255,15 @@ private void displayTokenIcon()
...
-            String tURL = assetDefinition.getTokenImageUrl(token.tokenInfo.chainId, token.getAddress());
-            if (TextUtils.isEmpty(tURL))
-            {
-                tURL = TRUST_ICON_REPO.replace(ICON_REPO_ADDRESS_TOKEN, correctedAddr);
-            }
-
-            boolean onlyTryCache = iconCheck.containsKey(correctedAddr);
+            IconItem iconItem = assetDefinition.fetchIconForToken(token);
 
             Glide.with(getContext().getApplicationContext())
```
JamesSmartCell, line 260: If we are going to use a new class for the iconItem then how about abstracting the code in .signature(...) out into the new class?

app/src/main/java/com/alphawallet/app/ui/TokenManagementActivity.java
```diff
@@ -64,15 +91,42 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         hideZeroBalanceCheckBox.setOnCheckedChangeListener((v, checked) -> {
             pref.edit().putBoolean("hide_zero_balance_tokens", checked).apply();
         });
```
JamesSmartCell, line 93: what does this do? Can you add a comment why we need the jake wharton library here?

Can we achieve the same thing without it using Runnable?

JamesSmartCell:
@ChintanRathod can you write why we need to add the new 'jake wharton' library? Not sure what it's doing - can we achieve the same using runnables; making the code more maintainable?

app/src/main/java/com/alphawallet/app/ui/TokenManagementActivity.java
ChintanRathod: The RX Binding library binds Search EditText directly and apply filterable to it. Hence we used it inside the code. Definitely we can remove the overburden of library and implement ourselves.

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokenListAdapter.java
ChintanRathod: I will make it work with single list

ChintanRathod:
@JamesSmartCell I did necessary changes as mentioned except the IconItem one. I didn't get what is expected.

JamesSmartCell:
@ChintanRathod One last thing can you update the 'confirm' button on this page to use FunctionButtonBar as per BackupKeyActivity etc.

JamesSmartCell:
Hi @ChintanRathod I think this can be done in a much simpler way which is more expandable.  For example in this PR we are maintaining two lists for showing and hidden. When this view is expanded we'll need to add new lists for each, which is prone to failure and code bloat, and would need complex interchange mechanisms - having nightmares trying to work on it.

I took a subset of this PR, created a new branch to get only the icons working in the add/hide tokens view. This branch is merged into master.

How about using the views to achieve the lists. So for example you can have three weighting 'zones' each starting from Integer.MAXIMUM / 3, for the three elements we'll eventually have. Each title can have exactly the weighting of that start point, and entries under the list can follow on the base weighting.

Handled like this the code would be have a much smaller footprint and be more maintainable and easier to expand. 

Can you start again from the current master, which has a subset of this PR implemented; you'll need to re-do the titles entries you added before but this time the code should be a lot less complex. In theory it should be possible to implement only touching a couple of files.

Also - needs to use FunctionButtonBar.

PR #1500, opened 2020-07-07
Camera features - Flash, My QR Code and Browse

### LINKED ISSUE #1495: Update Light, My QR Code, Browse buttons in Scan QR Code screen
**Current:**

**We need:** 
a) Light - tour on the light in your camera
b) My QR Code - switches to "My Wallet Address" screen
c) Browse - to select images from your Camera Roll

Screenshot 2020-07-03 10 36 30

ChintanRathod:
Fixes #1495

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/HomeActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/HomeActivity.java
@@ -945,16 +945,16 @@ public class HomeActivity extends BaseNavigationActivity implements View.OnClick
             case C.REQUEST_UNIVERSAL_SCAN:
                 if(resultCode == Activity.RESULT_OK)
                 {
-                    if(data.hasExtra(C.EXTRA_QR_CODE))
+                    if (data.hasExtra(C.EXTRA_QR_CODE))
                     {
                         String qrCode = data.getStringExtra(C.EXTRA_QR_CODE);
                         viewModel.handleQRCode(this, qrCode);
                     }
-                    else if(data.hasExtra(C.EXTRA_ACTION_NAME))
+                    else if (data.hasExtra(C.EXTRA_ACTION_NAME))
                     {
                         String action = data.getStringExtra(C.EXTRA_ACTION_NAME);
 
-                        if(action.equalsIgnoreCase(C.ACTION_MY_ADDRESS_SCREEN))
+                        if (action.equalsIgnoreCase(C.ACTION_MY_ADDRESS_SCREEN))
                         {
                             viewModel.showMyAddress(this);
                         }
--- a/app/src/main/java/com/alphawallet/app/ui/HomeActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/HomeActivity.java
@@ -945,8 +945,20 @@ public class HomeActivity extends BaseNavigationActivity implements View.OnClick
             case C.REQUEST_UNIVERSAL_SCAN:
                 if(resultCode == Activity.RESULT_OK)
                 {
-                    String qrCode = data.getStringExtra(C.EXTRA_QR_CODE);
-                    viewModel.handleQRCode(this, qrCode);
+                    if(data.hasExtra(C.EXTRA_QR_CODE))
+                    {
+                        String qrCode = data.getStringExtra(C.EXTRA_QR_CODE);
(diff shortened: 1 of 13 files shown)


ChintanRathod:
@JamesSmartCell This PR has all 3 buttons implementation. But there is an issue I am facing with Zxing library to decode bitmap we select from "Browse". I tried to fix an issue of Zxing library decode but not able to fix it. With this issue, after selecting image, we are getting "com.google.zxing.NotFoundException". I will check again to fix this issue.

JamesSmartCell:
> @JamesSmartCell This PR has all 3 buttons implementation. But there is an issue I am facing with Zxing library to decode bitmap we select from "Browse". I tried to fix an issue of Zxing library decode but not able to fix it. With this issue, after selecting image, we are getting "com.google.zxing.NotFoundException". I will check again to fix this issue.

I expected that would be an issue. I wasn't sure if the decode library would accept an image directly or not.

ChintanRathod:
I tried with different solutions like Resize, rotate and scaling. Yet not able to find the root cause.  
2 observations, 
- Screenshots are decoded successfully
- Camera pictures having exception

JamesSmartCell:
Can we lock the orientation while in this mode so we don't need to worry about it?

hboon:
@JamesSmartCell QR code from photo library: if it's tedious/buggy, might be worth just skipping the feature? It doesn't seem like an incredibly useful feature. I only use it because I normally develop for iOS in the simulator (which doesn't have a camera).

ChintanRathod:
I went through many Stackoverflow and blogs which states that Zxing library is built for scanning through preview and has no core functionality to decode captured image. Many people are facing same issue yet no solid solution I was able to find. 
Ref. 
https://github.com/journeyapps/zxing-android-embedded/issues/395

JamesSmartCell:
Removing the browse seems to be the best solution. Is the 'plumbing' for the browse still present or has it been removed too?

app/src/main/java/com/alphawallet/app/ui/HomeActivity.java
```diff
@@ -945,8 +945,20 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data)
             case C.REQUEST_UNIVERSAL_SCAN:
                 if(resultCode == Activity.RESULT_OK)
                 {
-                    String qrCode = data.getStringExtra(C.EXTRA_QR_CODE);
-                    viewModel.handleQRCode(this, qrCode);
+                    if(data.hasExtra(C.EXTRA_QR_CODE))
```
JamesSmartCell, line 948: Please use a space between if and the bracket

app/src/main/java/com/alphawallet/app/ui/QRScanning/CameraPreview.java
```diff
@@ -309,4 +309,31 @@ public void onAutoFocus(boolean success, Camera camera) {
     private void scheduleAutoFocus() {
         mAutoFocusHandler.postDelayed(doAutoFocus, 1000);
     }
+
+    public void turnOffFlash() {
+        try {
```
JamesSmartCell, line 314: inconsistent bracing - the preferred code standard for AW is to use a separate line for bracing eg:
```
if (a=b)
{
    do something;
}
```

app/src/main/java/com/alphawallet/app/ui/zxing/QRScanningActivity.java
```diff
@@ -118,4 +189,71 @@ public void onBackPressed()
...
+                        .subscribe(this::onSuccess, this::onError);
+            }
+        }
+    }
+
+    private void onError(Throwable throwable)
+    {
+        Toast.makeText(this, getString(R.string.error_browse_selection), Toast.LENGTH_LONG).show();
+    }
+
```
JamesSmartCell, line 214: The error should be handled by the calling function - usually a dialog box.

JamesSmartCell:
@ChintanRathod just a few formatting changes for some bad formatting that's crept in.

When handling the errors - can you check if these are handled by the activity/fragment that's calling this function? In most cases it should be handled there. So - if something is scanned and the QR code can't be classified and flow is passed back to the HomeActivity then the handling should be in there.

ChintanRathod:
@JamesSmartCell Understood.

PR #1509, opened 2020-07-14
Update "Add / Hide Tokens" screen 

### LINKED ISSUE #1487: Update "Add / Hide Tokens" screen

ChintanRathod:
Fixes #1487


ChintanRathod:
@JamesSmartCell In this, with respect to Zeplin design, we don't have "Apply Filter" function button, so I hide for the moment and handle the events accordingly. 
I suppose we are following modern events to immediate update without taking further confirmation from the user. 
Could you please review and let me know that is it good to go with?

PR #1510, opened 2020-07-15
Warning on sharing key backup

colourful-land:
add warning (ios do the same please)


AW-STJ:
The change does not allow the user to see the backup seed phrase. But the warning message is displayed

PR #1520, opened 2020-07-19
Activity infrastructure refactor

JamesSmartCell:
This large refactor accomplishes the following:

- UI fully decoupled from the blockchain queries and backend calculations etc. UI uses Realm listener to update view elements when they have been updated in the database. Values are stored in the database via the backend blockchain interface.
- "Single source of truth". UI always updates each element from the same data source.
- Create abstracted "meta" token cards. These contain enough information to disambiguate one token, and order it in a list view.
- Remove deprecated update systems and connections from Blockchain to UI info.

To see an example, see the before and after for WalletViewModel. You will see that now it consists of only an initial database fetch for the MetaTokenCards and some ancillary parts. In the UI, WalletFragment now uses a database listener to update the token cards.


James-Sangalli:
@JamesSmartCell this PR is massive and not atomic, not really sure how I can review it?

JamesSmartCell:
Note, this branch is not intended to be reviewed. It will become the new master once we have established that it's stable.

PR #1529, opened 2020-07-23
Add hide tokens new

### LINKED ISSUE #1487: Update "Add / Hide Tokens" screen

ChintanRathod:
Fixes #1487

### DIFF
--- a/app/src/main/java/com/alphawallet/app/repository/TokensRealmSource.java
+++ b/app/src/main/java/com/alphawallet/app/repository/TokensRealmSource.java
@@ -20,6 +20,7 @@ import com.alphawallet.app.repository.entity.RealmAuxData;
 import com.alphawallet.app.repository.entity.RealmERC721Asset;
 import com.alphawallet.app.repository.entity.RealmToken;
 import com.alphawallet.app.repository.entity.RealmTokenTicker;
+import com.alphawallet.app.repository.entity.RealmWalletData;
 import com.alphawallet.app.service.AssetDefinitionService;
 import com.alphawallet.app.service.RealmManager;
 
@@ -601,6 +602,13 @@ public class TokensRealmSource implements TokenLocalSource {
         return assets;
     }
 
+    /**
+     * Fetches all enabled TokenMetas in database, adding in chain tokens if required
+     * @param wallet
+     * @param networkFilters
+     * @param svs
+     * @return
+     */
     public Single<TokenCardMeta[]> fetchTokenMetas(Wallet wallet, List<Integer> networkFilters, AssetDefinitionService svs)
     {
         List<TokenCardMeta> tokenMetas = new ArrayList<>();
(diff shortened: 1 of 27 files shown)


app/src/main/java/com/alphawallet/app/repository/TokensRealmSource.java
```diff
@@ -595,23 +595,36 @@ private boolean traitsDifferent(List<Trait> traits, List<Trait> traits1)
         return assets;
     }
```
JamesSmartCell, line 597: Maybe create a new method to fetch all the TCMs (fetchAllTokenMetas); expanding the existing one is heading for trouble.

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokenListAdapter.java
```diff
@@ -170,13 +300,85 @@ public void onClick(View v) {
...
         void onItemClick(Token token, boolean enabled);
     }
+
+    @Override
+    public int getItemViewType(int position) {
+        return items.get(position).viewType;
+    }
+
+    @Override
+    public Filter getFilter() {
```
JamesSmartCell, line 336: Surely there's a simpler way to implement the search facility. So many flags and logic paths, it's impossible to work on. Can we use a different approach to achieve a more elegant method?

ChintanRathod: I understand your concern but the complexity of the list is its Type. I need to manage the Single list with 4 parameters. 
1. Display Label, 2. Display Tokens, 3. Hidden Label and 4. Hidden Tokens

Searching leads to show hide views internally and switching between list made it too complex in the scenario like When a list in "Display" state, switched off, has to transfer to "Hidden" state and list as well. Such scenario was bit tedious to manage without flags and complex logic.

JamesSmartCell: Yes, understood that you need those differences in views; I'm just a bit concerned that there's a very long method with a lot of logic branches in it. This can be solved by having a separate viewholder for each of the different views. For example the title views can be a very small local class; instantly you've simplified the logic, probably sufficiently to eliminate a lot of the different paths through the method.

Let's do this and see what the code looks like after that

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokenListAdapter.java
```diff
@@ -37,65 +42,188 @@
...
-        viewHolder.chainId = token.tokenInfo.chainId;
-        viewHolder.displayTokenIcon(token);
-        viewHolder.switchEnabled.setChecked(token.tokenInfo.isEnabled);
-        viewHolder.switchEnabled.setTag(new Integer(i));
-        viewHolder.switchEnabled.setOnCheckedChangeListener((v, b) -> {
-            listener.onItemClick(token, b);
-        });
-    }
+    public void onBindViewHolder(@NonNull RecyclerView.ViewHolder viewHolder, int position) {
+
```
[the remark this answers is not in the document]
ChintanRathod: Separate ViewHolders created with Binders.

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokenListAdapter.java
ChintanRathod: I applied changes for above 2 points for Separate method and ViewHolders. About the filter method, I felt that it is necessary logic and it is at simplest. No doubt it seems longer, but it has only 2 logic points. 
1) When search string is blank, mark visible all views
2) When search string is present, compare string with name and mark invisible which are not in the list.

I used this approach so that we don't need to take any other "backup" array list which can be used to fill adapter list again. By simply using a flag to display item in the list works well without creating and dropping items inside the adapter. 

Could you please share you thoughts on this?

PR #1531, opened 2020-07-24
Allow authentication via WalletConnect

justindg:
This PR allows authentication via WalletConnect on WC-enabled DApps (ie. https://localcryptos.com/login)

### DIFF
--- a/app/src/main/java/com/alphawallet/app/walletconnect/WCClient.kt
+++ b/app/src/main/java/com/alphawallet/app/walletconnect/WCClient.kt
@@ -1,0 +1,313 @@
+package com.alphawallet.app.walletconnect
+
+import android.util.Log
+import com.alphawallet.app.util.Utils
+import com.alphawallet.app.walletconnect.entity.*
+import com.alphawallet.app.walletconnect.util.WCCipher
+import com.alphawallet.app.walletconnect.util.toByteArray
+import com.github.salomonbrys.kotson.fromJson
+import com.github.salomonbrys.kotson.typeToken
+import com.google.gson.GsonBuilder
+import com.google.gson.JsonArray
+import okhttp3.*
+import okio.ByteString
+import java.util.*
+
+open class WCClient(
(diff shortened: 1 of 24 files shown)


app/src/main/java/com/alphawallet/app/walletconnect/WCClient.kt
```diff
@@ -0,0 +1,313 @@
...
+                }
+            }
+            WCMethod.ETH_SIGN -> {
+                val params = gson.fromJson<List<String>>(request.params)
+                if (params.size < 2)
+                    throw InvalidJsonRpcParamsException(request.id)
+                onEthSign(request.id, WCEthereumSignMessage(params, WCEthereumSignMessage.WCSignType.MESSAGE))
+            }
+            WCMethod.ETH_PERSONAL_SIGN -> {
+                val params = gson.fromJson<List<String>>(request.params)
```
JamesSmartCell, line 248: We are already including a gson library. Can we use the one we're already importing (via a Java call) rather than adding a duplicate library?

justindg: Are you referring to the kotson library? It is not a duplicate gson lib, just a lib containing helper/extension functions

JamesSmartCell: When I see unknown libraries which have people's names in the path I get nervous about them. What does it do? Is it an open source library, is it credible? I'll have to check the library out.

justindg: Ahh I see what you mean. No worries, I'll remove it and use the regular way!

JamesSmartCell: I checked it out it looks legit with a lot of contributors and is mainly for code clarity. If you haven't started the change then keep it as it is.

PR #1535, opened 2020-07-27
Protecting Backup and Import screen

### LINKED ISSUE #1514: αW not protecting the wallet recovery user entry screen
~~1. When the user enters backup-activity, he gets a message explaining what's seed phrase / keystore json. at which point he can't take a screenshot or make video recording. This screen *should not* be protected (the one after it should)~~ (removed as Chintan raised a question on this and I prefer to progress the next item without addressing it)

2. When the user recover from a backup, he types json keystore and this screen can be video recorded - where it *should* be protected.

ChintanRathod:
Fixes #1514


ChintanRathod:
@JamesSmartCell I thought to add "SECURE_FLAG" in ImportWallet screen too. 
Could you please check if that was not necessary, I will remove the flag and change the PR.

JamesSmartCell:
The idea for this - as I understand it - is to allow users to photograph the initial backup screen, but not to backup the rest. The design envisaged was for a separate, tiny activity for the initial screen then it goes into the backupkeyactivity, at one of two entry points depending on if you are backing up a seed phrase or a keystore.

I guess this solution might be ok but because it's still all one activity it's going to have to be carefully analysed to be sure there's no security issue. With two separate activities it's impossible to have a security issue and the code is cleaner. Please look at the issue again and let me know if it makes sense or not.

ChintanRathod:
In my opinion, for current requirement, the solution will work as I put secure flag at below 2 points
- setupTestSeed (while displaying Seed Phrase) 
- jumbleList (when displaying Seed Phrase in random order)
That's correct, if we create separate activity for such secure flow but I tested and out of 4 steps 2 steps are protected as required. So till we have separate methods to display contents, it won't create a problem for us.

JamesSmartCell:
I will break the required implementation down further as I think it's still not clear.

1. On user clicking 'Back up this wallet' user is taken to a new activity.

This activity does only two things:
      1. Show the 'What is Keystore JSON'
      2. Show the 'Back up your wallet with seed phrase'

If 1. then display button with 'back up my wallet', if 2 then display 'Export keystore JSON' button.

When button is clicked, open new activity which is screen locked. This activity will be the existing 'BackupKeyActivity', which is prevented from having picture taken.

2. On button being clicked, if 1. then take user to the 'Set Keystore Password', if 2. take user to 'Write down seed phrase on paper'.

The reason to do it this way is to prevent any future change from have unforeseen consequences. The current PR still does not match the issue request.

ChintanRathod:
@JamesSmartCell Could you please review the new activity and the flow if it is not breaking anywhere else. I tested from my side but yet to be confirm with the smooth flow like previous it has.

PR #1554, opened 2020-08-04
Add a universal token icon class

### LINKED ISSUE #1550: Create a universal token icon class
This new class should appear similarly to the 'FunctionButtonBar' etc.

Should be a new type that has a layout and attached class that you can drop into any view and populate from that view.

The icon should contain the TextView 'textIcon' as well as the ImageView 'icon' in the current item_token.xml

```
            <TextView
                android:id="@+id/text_icon"
                android:layout_width="40dp"
                android:layout_height="40dp"
                android:layout_marginEnd="16dp"
                android:background="@drawable/grid_icon"
                android:backgroundTint="@color/azure"
                android:fontFamily="@font/font_bold"
                android:gravity="center"
                android:layout_gravity="center"
                android:lines="1"
                android:text="UEFA"
                android:textColor="@color/white"
                android:visibility="gone"
                app:autoSizeMaxTextSize="@dimen/sp13"
                app:autoSizeMinTextSize="@dimen/sp12"
                app:autoSizeStepGranularity="1sp"
                app:autoSizeTextType="uniform" />

            <ImageView
                android:id="@+id/icon"
                android:layout_width="40dp"
                android:layout_height="40dp"
                android:layout_marginEnd="16dp"
                android:contentDescription="@string/empty"
                android:gravity="center"
                android:layout_gravity="center"
                android:foreground="@drawable/white_circle"
                android:background="@color/logo_grey"
                android:src="@drawable/ropsten_logo"
                android:visibility="visible"
                />
```

So I'd drop in a view like this:

```
        <com.alphawallet.app.widget.TokenIcon
            android:id="@+id/token_icon"
            android:layout_width="wrap_content"
            android:layout_height="wrap_content" />
```

Anyone using this class will have to assign the AssetDefinitionService to the underlying class, and the class would contain all the code in the current 'TokenHolder' icon resolution code.

This will deduplicate a lot of code as we're now using this icon resolution in a lot of places.

Then, replace the (currently two) uses - class TokenHolder and class TokenListAdapter to use this new view container.

Note that in class TokenHolder the glide error 'URL not found' is blanked by this code:

```
    /**
     * Prevent glide dumping log errors - it is expected that load will fail
     */
    private RequestListener<Drawable> requestListener = new RequestListener<Drawable>() {
        @Override
        public boolean onLoadFailed(@Nullable GlideException e, Object model, Target<Drawable> target, boolean isFirstResource) {
            return false;
        }

        @Override
        public boolean onResourceReady(Drawable resource, Object model, Target<Drawable> target, DataSource dataSource, boolean isFirstResource) {
            return false;
        }
    };
```

There is a second part of this Issue as well, we need to be able to place a second smaller status icon at the bottom right of the main icon.

See the zeplin here:

https://zpl.io/bJR5gG3

Notice the icons at the bottom right overlaid on top of the main icon.

You will have to look through the zeplin layouts for the 4 currently known status SVG images (or check the zeplin resources):

- Receive (downward arrow in green circle)
- Send (diagonal upward arrow in green circle)
- Pending ('clock face' in grey circle)
- Failed (white 'x' in red circle)

plus, one more which isn't in zeplin assets:

- Transaction dropped.

I will produce a placeholder icon for this one

Also we should allow a custom image to be placed in there - this would be scaled to the same size.

Let me know if you need any more clarification, this work is required for the Activity upgrade.

ChintanRathod:
Fixes #1550

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/widget/entity/StatusType.java
+++ b/app/src/main/java/com/alphawallet/app/ui/widget/entity/StatusType.java
@@ -1,0 +1,9 @@
+package com.alphawallet.app.ui.widget.entity;
+
+/**
+ * Created by JB on 5/08/2020.
+ */
+public enum StatusType
+{
+    SENT, RECEIVE, PENDING, FAILED, REJECTED
+}
--- a/app/src/main/java/com/alphawallet/app/ui/widget/holder/TokenHolder.java
+++ b/app/src/main/java/com/alphawallet/app/ui/widget/holder/TokenHolder.java
@@ -1,8 +1,6 @@
 package com.alphawallet.app.ui.widget.holder;
 
 import android.graphics.Color;
-import android.graphics.drawable.BitmapDrawable;
-import android.graphics.drawable.Drawable;
 import android.os.Bundle;
 import android.os.Handler;
 import android.support.annotation.NonNull;
@@ -15,7 +13,6 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.view.animation.AlphaAnimation;
 import android.view.animation.Animation;
-import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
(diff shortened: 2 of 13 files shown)


JamesSmartCell:
Hi @ChintanRathod here's the icon code for 'transaction rejected':

```
<vector xmlns:android="http://schemas.android.com/apk/res/android"
    android:width="24dp"
    android:height="24dp"
    android:viewportWidth="24"
    android:viewportHeight="24">
  <path
      android:pathData="M12,12m-12,0a12,12 0,1 1,24 0a12,12 0,1 1,-24 0"
      android:fillColor="@color/warning_red"
      android:fillType="evenOdd"/>
  <path android:fillColor="@android:color/white" android:pathData="M3,8.41l9,9l7,-7V15h2V7h-8v2h4.59L12,14.59L4.41,7L3,8.41z"/>
</vector>
```

ChintanRathod:
@JamesSmartCell Icon added. Could you please review now.

PR #1563, opened 2020-08-10
Add support of Biometric authentication

### LINKED ISSUE #1555: Upgrade Fingerprint authentication to use the new classes
Currently we use the FingerprintManager class to interface with the device's biometric unlock code. This code has been deprecated and may have already started to cause issues.

Upgrade the code in class SignTransactionDialog to use the new class BiometricManager.

See the Google blog here:

https://source.android.com/security/biometric

and relevant StackOverflow info here:

https://stackoverflow.com/questions/51375606/use-fingerprint-is-deprecated-in-api-level-28

ChintanRathod:
Fixes #1555


ChintanRathod:
@JamesSmartCell I tested this Biometric authentication with Emulator Fingersensor but yet to validate with Face detection. Could you please check if we can do that with anyone's available device as don't have physical device here.

ChintanRathod:
BiometricManager is supported with Androidx and I tried with the AW current implemntation, though it is creating some issue that our whole project is not supporting with Andoridx yet. So I tried to implement with best possible option till date available. 

Please let me know if we need to work on this approach and make it more useful with advance implementation.

JamesSmartCell:
@ChintanRathod after looking at this we should try to keep the updated system unlock screen and allow user to use PIN if they fail with the print too many times. 

Can you make the following update:

- If the user fails to make the print, that is you end up at ```FINGERPRINT_ERROR_LOCKOUT``` or ```FINGERPRINT_ERROR_LOCKOUT_PERMANENT``` or ```FINGERPRINT_ERROR_UNABLE_TO_PROCESS```
then we should offer the user the chance to enter their PIN. This can be the old screen with 'cancel' on the left and 'use pin' on the right.

ChintanRathod:
@JamesSmartCell Could you please validate the BioMetric implementation once it got error with mentioned ERROR CODE. It will show PIN dialog.

JamesSmartCell (approved):

PR #1569, opened 2020-08-18
Update strings.xml

bzadik:
Added localizations by Babble-on (www.ibabbleon.com)


zhangzhongnan928 (commented):
I have reviewed Chinese translation without in app context. And I made some small changes.
Once it merges into app, I can review again inside the app.

James-Sangalli (approved):

PR #1571, opened 2020-08-20
Trivial issues on Browser URL bar

### LINKED ISSUE #1549: trivial issues on browser url bar
this is more of a improvement, not a serious issue but impact on the user experience a lot,
when I use the browser module, each time I want to search something, I'll tap on the url bar,
and just like the screenshots in below, sometimes it will expand a bit, still ok for me to click on the 'close' icon to clear the url,
but sometimes it won't expand, so I have to scroll to the end to find the 'close' icon,
image

image

based on experience on other similar browser app, my suggestion is that it will be better if you  can just clear out the url bar automatically, if user want to copy the url, you can put it in the dotdotdot tray

ChintanRathod:
Fixes #1549


JamesSmartCell:
Looks like the 'easy' way to do this would be to extend the AutoCompleteTextView. It's a bit of a can of worms.

ChintanRathod:
Sure let me check for chrome open source code.

JamesSmartCell:
Hi @ChintanRathod,

I cherry picked and created a new branch from master based on your commits here: https://github.com/AlphaWallet/alpha-wallet-android/tree/fix-browser-bar-2

It's really close! This one is works perfectly but still has the vibrate after the expand. Is there any way to stop the vibrate? If so this PR can go through.

ChintanRathod:
@JamesSmartCell Let me check for this buzz issue. I didn't implemented anywhere but it seems somewhere OS automatically activates it. I will check and let you know.

PR #1579, opened 2020-08-28
Protecting Backup and Import screen

### LINKED ISSUE #1514: αW not protecting the wallet recovery user entry screen

ChintanRathod:
Fixes #1514


JamesSmartCell:
@ChintanRathod Crash bug: 

1. Create a new wallet.
2. Click on backup the wallet, and click on the 12 words in the correct order.
3. Where you get the 'upgrade my key', click back (Flow error: you should return to the settings page, not go to the intermediate page).
4. Click back again (see flow error above - should only need to click 'back' once), to go to the settings page.
5. From settings page, click on 'Upgrade key security' - crash.

Once you get to BackupKeyActivity the BackupFlowActivity should terminate, so that there's no intermediate screen.

ChintanRathod:
@JamesSmartCell Ok. Got the flow. I will check and make the appropriate flow fix.

JamesSmartCell (approved):

PR #1581, opened 2020-08-30
Fix for default and override locale setting

JamesSmartCell:
This PR fixes a few issues noticed with the locale settings, which has become a bit messy after a few re-edits.

Issues: 

1. When phone was switched to a different locale (without user explicitly choosing a locale from the settings), the phone wouldn't always pick up the changed locale.
2. Text in some menus didn't appear in the locale language on restarting the app.
3. Duplicated code.
4. Some translations were missed (see values-zh/strings.xml)

While editing the relevant classes, cleanup of redundant code was also done.

Code was checked on Pixel, Xiaomi and Huawei phones set to different locales. To repeat a check:

1. Create new emulator instance.
2. Set device language to Chinese or Spanish
3. Install AW - text should appear in the device language.
4. Change language - should be in the override language.
5. Restart app, setting should persist.
Bonus: Check the top right hamburger menu in DappBrowser; previously text was in English no matter what language settings were applied.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/di/NewSettingsModule.java
+++ b/app/src/main/java/com/alphawallet/app/di/NewSettingsModule.java
@@ -2,17 +2,11 @@ package com.alphawallet.app.di;
 
 import com.alphawallet.app.interact.GenericWalletInteract;
 import com.alphawallet.app.interact.GetDefaultWalletBalance;
-
-import com.alphawallet.app.repository.CurrencyRepository;
-import com.alphawallet.app.repository.CurrencyRepositoryType;
 import com.alphawallet.app.repository.EthereumNetworkRepositoryType;
-import com.alphawallet.app.repository.LocaleRepository;
-import com.alphawallet.app.repository.LocaleRepositoryType;
 import com.alphawallet.app.repository.PreferenceRepositoryType;
 import com.alphawallet.app.repository.WalletRepositoryType;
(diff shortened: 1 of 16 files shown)


James-Sangalli (changes requested):
@JamesSmartCell if you kill the app it resets back to English, otherwise it works as expected.

JamesSmartCell:
> @JamesSmartCell if you kill the app it resets back to English, otherwise it works as expected.

Unable to reproduce. Can you describe how to reproduce in more detail?

James-Sangalli:
> > @JamesSmartCell if you kill the app it resets back to English, otherwise it works as expected.
> 
> Unable to reproduce. Can you describe how to reproduce in more detail?

On the emu I change the language, kill the app and then reopen it.

I expect it to be in the changed language but it reverts back to English instead

JamesSmartCell:
@James-Sangalli fresh install or old version it works correctly for me on all test phones. How about I give you a build, perhaps there was a build issue. I rebased too. I'll be preparing a release candidate today, so you can test that.

JamesSmartCell:
Here is a new build; please check if it works on your device as expected:

https://drive.google.com/file/d/17j28cRWduOGHqSydPK7ieHtVhZ4YljPe/view?usp=sharing

JamesSmartCell:
Merging this PR as verified correctly.

PR #1586, opened 2020-09-03
Improved search with Popular tokens

### LINKED ISSUE #1570: Improve search results and add Top 20 Most Popular tokens
**Problem:**
During user tests, I discover that users have problems understanding how the wallet works. They think if something is not listed in the wallet tab - it is not available. And even if they try to look for a specific coin in the Add/Hide Tokens - if they can't search or find tokens they want to transfer to the wallet - they assume these tokens are not available for this wallet.

One user thought that a coin is not available because it was not listed in the wallet tab. So there was a wrong assumption that if something was not visible, it must be unavailable.

**Solution:**

1. Below "Hidden Tokens" add another section called "Popular Tokens" where we list the top 20 most popular erc20 tokens.
https://zpl.io/VKMMAjX
2. The Search box in Add/Hide Tokens should display (as a result of searching) even those tokens that are out of the wallet, but still, ERC20.
https://zpl.io/VKMMAjX https://zpl.io/aNXXGjn
3. Add Search bar (working the same way as 2) to the Add Custom Token screen.
https://zpl.io/beRRBZP

Not related but can be updated along with this change:
4. Remove Plus icon symbol from the wallet (leave just a text "Add/Hide Tokens"
https://zpl.io/aRdOwWK
5. Make sure that the change from https://github.com/AlphaWallet/alpha-wallet-android/issues/1487 is also applied.
6. You can remove "Hide Zero Balance Tokens" Feature

Add_Hide_Tokens_Android

ChintanRathod:
Fixes #1570


app/src/main/java/com/alphawallet/app/service/TokensService.java
```diff
@@ -112,7 +112,7 @@ private void checkUnknownTokens()
```
JamesSmartCell, line 112: TokensService should not know about 'isEnabled', it is a backend service and should be 100% decoupled from UI. This need to be refactored. See above comment about 'isEnabled' not being in the ContractLocator class.

lib/src/main/java/com/alphawallet/token/entity/ContractAddress.java
```diff
@@ -10,17 +10,27 @@
 {
     public final int chainId;
```
JamesSmartCell, line 11: isEnabled needs to be removed from this class. It should only be in the TokenMeta

app/src/main/java/com/alphawallet/app/repository/TokenRepository.java
```diff
@@ -1222,7 +1227,25 @@ private String callCustomNetSmartContractFunction(
                     getName(address, network),
                     getContractData(network, address, stringParam("symbol"), ""),
                     getDecimals(address, network),
-                    true, chainId);
+                    true,
+                    chainId);
+        });
```
JamesSmartCell, line 1232: Remove setupTokensFromLocal.

I will further clarify the flow in a longer comment at the end of the review.

JamesSmartCell (changes requested):
The expected flow should be like this:

1. Read in the popular tokens.
2. Feed popular tokens into TokenService method addUnknownTokenToCheck
3. Create TokenMeta entries for all the popular tokens (that we just read) in the list; note that at this stage not all tokens may be resolved, so at least the first time this method is clicked we can expect a little item view 'Jank'.
4. As the tokens become available populate the views in the list. To achieve this you will need a database listener as per WalletFragment -> TokensAdapter (see setRealmListener).

Following this system; the backend is decoupled from the frontend and you can see a lot of the 'roots' of the code are no longer needed - the number of files drops and the classes become a lot more self contained.

The large change that just happened in the wallet was to remove the coupling between the UI and the back end. In this manner a lot less code needs to be added for each change. The old methods of doing things lead to sloppy difficult to maintain code that has hooks all through the wallet.

Let me know if you get stuck anywhere or if any of the points aren't clear! It's a bit of a change from the way it was done before but it's worth to change to this system.

Eventually I'd like to completely decouple the backend into a separate module or library that can be used for other projects.

JamesSmartCell:
@ChintanRathod More information and a slight change in direction:

These tokens should be included by default, so if a user is importing a wallet we should be picking up these tokens in the whole wallet not just in the management screen.

To do this, look at how the 'known_contracts,xml' resource is handled. This is scanned when the wallet first starts.

This way, the tokens should be available when the user first enters the management page.

The size of this project should then very small, since it only matching how the 'known_contracts.xml' functionality. Still go down the route of the database listener so you can match the way that the other adapters work, and if the user clicks on the management page before the new contracts list is complete, the tokens will be added as they are discovered.

app/src/main/java/com/alphawallet/app/entity/ContractLocator.java
```diff
@@ -20,26 +20,38 @@
...
     }
 
     public ContractLocator(String n, int chain, ContractType t)
     {
         address = n;
         chainId = chain;
         type = t;
+        isEnabled = true;
+    }
+
```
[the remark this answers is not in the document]
ChintanRathod: Removed

app/src/main/java/com/alphawallet/app/repository/TokenRepository.java
ChintanRathod: removed

lib/src/main/java/com/alphawallet/token/entity/ContractAddress.java
ChintanRathod: removed

app/src/main/java/com/alphawallet/app/service/TokensService.java
ChintanRathod: removed

JamesSmartCell (approved):

PR #1602, opened 2020-09-14
Add analytics and noAnalytis flavor

### LINKED ISSUE #1591: Complete MixPanel setup
Now that flavours have been re-done, it should be possible to do a clean 'No Firebase' flavour for the China release - this shouldn't require any code duplication this time.

ChintanRathod:
Fixes #1591


ChintanRathod:
@JamesSmartCell I implemented for Firebase analytics as in this project it is only Firebase integrated. Let me know if we also need to take care of MixPanel to alphawallet.

JamesSmartCell (approved):

PR #1605, opened 2020-09-15
"Setting Subtitle" message should be a contract address or a few contact addresses

### LINKED ISSUE #1599: "Setting Subtitle" message should be a contract address or a few contact addresses
1. Enable TS debugging.

2. Drop this file to the AlphaWallet directory on the device;
https://github.com/AlphaWallet/TokenScript/blob/access-cards-from-web/tests/EntryToken.xml

3. Observe in "TokenScript Management" there is "Setting Subtitle"

Screenshot_20200914-185512_AlphaWallet

Expected:

The newly added TokenScript should look like the others.

ChintanRathod:
Fixes #1599

### DIFF
--- a/app/src/main/res/layout/item_tokenscript_management.xml
+++ b/app/src/main/res/layout/item_tokenscript_management.xml
@@ -81,7 +81,7 @@
             android:fontFamily="@font/font_light"
             android:lines="1"
             android:maxLines="1"
-            android:text="Setting Subtitle"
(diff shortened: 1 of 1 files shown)


ChintanRathod:
@colourful-land 
The issue was due to hardcoded "Setting Subtitle" text. It will set address once it is able to fetch from the file as like below. 

image

JamesSmartCell (approved):

PR #1629, opened 2020-10-06
Fix for issues where Iris/Face unlock fails for whatever reason.

### LINKED ISSUE #1622: Facial ID not working for Samsung Devices
From my chat with @colourful-land, looks like the facial ID for signing transactions does not work on Samsung devices. 

@JamesSmartCell Would you be able to add any details relevant to the issue?

JamesSmartCell:
Fixes #1622

Fallback to system unlock prompt if biometric unlock fails.

Note - many developers report that Samsung face/iris unlock is not completely stable, so the fallback option ensures that there is always a way to unlock the key.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/service/KeyService.java
+++ b/app/src/main/java/com/alphawallet/app/service/KeyService.java
@@ -616,10 +616,10 @@ public class KeyService implements AuthenticationCallback, PinAuthenticationCall
             boolean success = writeBytesToFile(ivPath, iv);
             if (!success)
             {
-                deleteKey(fileName);
+                //deleteKey(fileName);
                 throw new ServiceErrorException(
                         ServiceErrorException.ServiceErrorCode.FAIL_TO_SAVE_IV_FILE,
-                        "Failed to saveTokens the iv file for: " + fileName + "iv");
+                        "Failed to create the iv file for: " + fileName + "iv");
             }
 
             try (CipherOutputStream cipherOutputStream = new CipherOutputStream(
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/alphawallet/app/service/KeyService.java
```diff
@@ -616,10 +616,10 @@ private synchronized boolean storeEncryptedBytes(byte[] data, boolean createAuth
             boolean success = writeBytesToFile(ivPath, iv);
             if (!success)
             {
-                deleteKey(fileName);
+                //deleteKey(fileName);
```
James-Sangalli, line 619: @JamesSmartCell better to simply delete the code rather than keep it round

app/src/main/java/com/alphawallet/app/widget/SignTransactionDialog.java
```diff
@@ -72,42 +76,64 @@ public SignTransactionDialog(@NonNull Activity activity, Operation callBackId, S
...
+            showAuthenticationScreen();
         }
         else
         {
-            getLegacyAuthentication(authCallback);
+            authCallback.authenticateFail("Device unlocked", AuthenticationFailType.DEVICE_NOT_SECURE, callBackId);
         }
     }
 
+//    //get fingerprint or PIN
```
James-Sangalli, line 103: Delete @JamesSmartCell

James-Sangalli:
@JamesSmartCell I will test this later from my samsung s8

app/src/main/java/com/alphawallet/app/service/KeyService.java
JamesSmartCell: There's a reason why I did this - 

with biometrics on Samsung phone certain biometric unlock doesn't result in a 'secure unlock' event, only a 'succeeded in proving face biometric' so the storeEncryptedBytes would fail if upgrading.

In this case we definitely want the key to remain otherwise the user has lost their key.

Since we can't completely rule out this issue in the future it's best to err on the side of caution and not delete keys if there's an issue with storing the key in the case where we're upgrading the key.

JamesSmartCell:
Received update from original (external) reporter that this fixes the issue.

James-Sangalli:
@JamesSmartCell so I tested on my galaxy s8. This app doesn't crash but it doesn't inform you that you must use the pin (it still says that you can use your iris)

JamesSmartCell:
Closing because out of date, replaced with newer fix

PR #1631, opened 2020-10-06
Address not fully populated in 'My Address' screen

### LINKED ISSUE #1619: Address not fully populated in 'My Address' screen
Last digit for the address is not populated. In the blow example, the last digit is '2' and that looks truncated. The QR code is populated accurately.

image

ChintanRathod:
Fixes  #1619

### DIFF
--- a/app/src/main/res/layout/item_copy_textview.xml
+++ b/app/src/main/res/layout/item_copy_textview.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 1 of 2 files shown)


colourful-land:
Can we simply drop the MaxWidth? You can never be sure how wide it is necessary because the user might be using a large system (accessibility option for older people)

ChintanRathod:
> Can we simply drop the MaxWidth? You can never be sure how wide it is necessary because the user might be using a large system (accessibility option for older people)

Sure we can remove it. I just implemented to make it more visually clear.

ChintanRathod:
@colourful-land Could you please again. I set the view properly after removing "maxwidth" property.

colourful-land (approved):
I assume that you tried to run it on your mobile and it works correctly there?

PR #1644, opened 2020-10-19
Upgrade to AndroidX

JamesSmartCell:
Long overdue upgrade to AndroidX.

Note that I had to attend to some details, mainly still using an older version of Gradle to keep the build size small. When upgrading to a later version of gradle I noticed a very large increase in release build size. Proguard instructions should be the solution however the documentation for Proguard is somewhat sparse and usage seems to be a kind of 'black magic' that's a closely guarded secret as there is a cottage industry of competing 3rd party versions (on 'hostage' subscriptions - that is, cheap for a year to get you hooked then massive increase to a sizable sum a year later and you'll pay because no-one has the time to go diving into proguard for a couple of days to solve it correctly).

Tested debug and release builds.

This needs to be merged into master ASAP because we need to test, test, test and have all current branches using this code to check for issues. I have payed with all the features and it seems ok.


James-Sangalli (approved):
Please squash your commits into one

PR #1652, opened 2020-10-26
Add token transfer events from Etherscan

JamesSmartCell:
Uses the Etherscan APIs to discover token transfers into and out of the wallet. Previously we discovered the transfers going out of the wallet, but with this API route when any token is transferred to the wallet, the receive event will appear and the token will always appear in the wallet.

NOTE: This PR builds on PR #1646 . PR #1646 needs to be pushed first.

### DIFF
--- a/app/src/androidTest/java/com/alphawallet/app/WalletTest.java
+++ b/app/src/androidTest/java/com/alphawallet/app/WalletTest.java
@@ -1,6 +1,6 @@
 package com.alphawallet.app;
 
-import androidx.test.ext.junit.runners.AndroidJUnit4;
+import androidx.test.runner.AndroidJUnit4;
 
 
 import org.junit.runner.RunWith;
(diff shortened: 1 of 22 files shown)


hboon:
@JamesSmartCell could you share briefly which Etherscan endpoint (new or old) and what what we extract from it?

JamesSmartCell:
Hi @hboon I believe @James-Sangalli has already implemented it on the iOS app.

```https://api.etherscan.io/api?module=account&action=tokentx&address=0x4e83362442b8d1bec281594cea3050c8eb01311c&page=1&offset=100&sort=asc&apikey=YourApiKeyToken```

Gives you all ERC20 sends and receives concerning the given address. The receives are really useful, since the sends we know about from the transaction log.

There is also an endpoint for ERC721 transfers which looks almost the same (notice the ```action=tokennfttx```):

```https://api.etherscan.io/api?module=account&action=tokennfttx&address=0x4e83362442b8d1bec281594cea3050c8eb01311c&page=1&offset=100&sort=asc&apikey=YourApiKeyToken```

hboon:
Ah right. I think we have this for iOS. Thanks.

James-Sangalli (approved):
Nice

Issue #1647, opened 2020-10-20
Android X Upgrade Testing Results


AW-STJ:
1.  FaceID available in Samsung and causes key import error
![image](https://user-images.githubusercontent.com/61957841/96582898-a7838a00-1327-11eb-94b5-e717424f4d7a.png)

2. 'Wallet Address QR Code' not working
![image](https://user-images.githubusercontent.com/61957841/96582969-c255fe80-1327-11eb-9ddd-12c77faf3ad2.png)

3. Wallet address font colour is white
![image](https://user-images.githubusercontent.com/61957841/96583029-d7329200-1327-11eb-8fb0-0eaf215f6276.png)

4. Incorrect token balance while loading the wallet for the first time
![image](https://user-images.githubusercontent.com/61957841/96583099-ed405280-1327-11eb-9b20-e4893e4e2051.png)

5. Long loading times for the wallet 
![image](https://user-images.githubusercontent.com/61957841/96583142-fd583200-1327-11eb-83f5-d024d8c11f28.png)

6. No receive transactions (probably ERC20s ) on other networks [Issue already raised, but clarifying its still in this build]
![image](https://user-images.githubusercontent.com/61957841/96583763-f2ea6800-1328-11eb-9d94-f618cac0e8b1.png)

7. Approve transaction not captured correctly
![image](https://user-images.githubusercontent.com/61957841/96583825-085f9200-1329-11eb-8035-d99e5ef4bb61.png)

JamesSmartCell:
Most of these are already fixed in the outstanding PR's. If they are still not approved I'll use supervisor push during the week, as we need them in the main branch for testing.

AW-STJ:
> 1.  FaceID available in Samsung and causes key import error
> 
Fixed in 3.21.0 (132)

> 2. 'Wallet Address QR Code' not working

Still open 

> 3. Wallet address font colour is white
Fixed in 3.21.0 (132)

> 4. Incorrect token balance while loading the wallet for the first time
Fixed in 3.21.0 (132)

> 5. Long loading times for the wallet 
Fixed in 3.21.0 (132)


> 6. No receive transactions (probably ERC20s ) on other networks [Issue already raised, but clarifying its still in this build]
Fixed in 3.21.0 (132)
> 
> 7. Approve transaction not captured correctly
Fixed in 3.21.0 (132)

JamesSmartCell:
@AW-STJ can this be closed now? All the above points are fixed, please check the latest build.

PR #1653, opened 2020-10-26
Fix cosmetic issue with 'Approve' values

JamesSmartCell:
App was displaying the actual approve value, like 99999999999999.9999 etc. Replace with heuristic that if the (abs) Approve value is more than 10x the token balance then show the approve as 'All' to make the display neater.

Also, fix issue in case there is a very long number displaying that number properly centred in the 'Token Activity' page (and tidy up the formatting in the file - the actual change is very small).

This fixes one of the issues seen in the image chain in issue #1647.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/TransactionContract.java
+++ b/app/src/main/java/com/alphawallet/app/entity/TransactionContract.java
@@ -13,6 +13,7 @@ import com.alphawallet.app.util.BalanceUtils;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.MathContext;
 import java.util.List;
 
 import static com.alphawallet.app.entity.TransactionOperation.NORMAL_CONTRACT_TYPE;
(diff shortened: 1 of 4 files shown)


James-Sangalli:
@JamesSmartCell can just compare against 0xFFF...FFF as that is the max number and the default approval value when approving 'ALL' no need to check against the user's balance and this could be error prone in the case that a user approves a smaller amount but their balance is under 10X.

app/src/main/java/com/alphawallet/app/entity/TransactionContract.java
```diff
@@ -176,8 +176,8 @@ public String getOperationResult(Token token, TransactionOperation operation, Tr
             if (!token.isNonFungible() && (operation.value == null || operation.value.length() > 0 && Character.isDigit(operation.value.charAt(0))))
             {
                 BigDecimal value = new BigDecimal(operation.value);
-                //appears to be a number; try to produce number with prefix
-                return tx.getPrefix(token) + BalanceUtils.getScaledValueFixed(value, token.tokenInfo.decimals, TransactionHolder.TRANSACTION_BALANCE_PRECISION);
+                if (getOperationId() == TransactionType.APPROVE.ordinal() && value.abs().compareTo(token.balance.multiply(BigDecimal.TEN)) > 0) return "All";
```
James-Sangalli, line 179: Compare to 0xFFF....FFF here instead for simplicity

JamesSmartCell: Sometimes the approve amount is not BigInteger(-1), in which case this won't work.

I've seen a few dapps use a different, large amount.

James-Sangalli (approved):

PR #1655, opened 2020-10-28
refactor WalletConnectActivity: use constant instead of hard coded value

theqp:


JamesSmartCell (approved):
The is the right way to do it.

PR #1656, opened 2020-10-29
fix NPE in Web3Transaction

theqp:
Reproducing the bug:
1. Connect with MyEtherWallet via WalletConnect
2. Try to send 0 ERC20 tokens

The app will go back to the main screen without any error displayed on the screen. The value of `wcTx.getValue()` is `0x` and `Hex.hexToBigInteger("0x")` returns `null`.


JamesSmartCell (approved):
Nice catch, thanks for the commit :)

PR #1664, opened 2020-11-04
Fix address QR format

JamesSmartCell:
This PR fixes the 'My Address' layout issue where the QR code has the top and bottom clipped due to constraint layout issues.


James-Sangalli (approved):
Tested with nexus emu

PR #1666, opened 2020-11-09
Fix unreliable Token name/decimals override

JamesSmartCell:
Fix for issue where token fetch from node is unreliable; if we have a result from Etherscan then use this.

Observe this token:

https://etherscan.io/token/0xe4c94d45f7aef7018a5d66f44af780ec6023378e

If you translate the 'name' from bytes:

0x70726f7879434300000000000000000000000000000000000000000000000000 bytes32

You'll see 'proxyCC', whereas the name of the token is CCRB as registered on Etherscan.

Also there's no decimals read, but the Etherscan info shows 6.

So, if we get a token return from Etherscan from a token transaction, the information from this should override the usual data.

This is a bad precedence, but not sure how else we can handle this.

To reproduce and check on iOS try watching this address:

```0x9c84EC90EF46e68530893cF3229A1Dd695b8b457```

### DIFF
--- a/app/src/main/java/com/alphawallet/app/repository/TokensRealmSource.java
+++ b/app/src/main/java/com/alphawallet/app/repository/TokensRealmSource.java
@@ -419,12 +419,7 @@ public class TokensRealmSource implements TokenLocalSource {
                 TransactionsRealmCache.addRealm();
                 realm.beginTransaction();
                 token.setRealmBalance(realmToken);
-                if (token.tokenInfo.name.length() > 0)
-                {
-                    realmToken.setName(token.tokenInfo.name);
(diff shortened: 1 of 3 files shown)


colourful-land:
> This is a bad precedence, but not sure how else we can handle this.

We probably should use TokenScript for this instead of etherscan but I agree the need of a quick fix.

JamesSmartCell:
> > This is a bad precedence, but not sure how else we can handle this.
> 
> We probably should use TokenScript for this instead of etherscan but I agree the need of a quick fix.

Agree completely - do you think the easiest way would be to crawl etherscan for tokens using a script and procedurally generate the scripts from a template, then upload them to the repo. These scripts would then be customisable by the relevant token owners as they see fit.

PR #1668, opened 2020-11-10
Add wallet name

### LINKED ISSUE #1665: Allow to name wallets
We have implemented the same functionality for Qoin. Let us get that into AlphaWallet as well.

ChintanRathod:
Fixes #1665

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/WalletActionsActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/WalletActionsActivity.java
@@ -12,6 +12,7 @@ import androidx.annotation.Nullable;
 import android.text.TextUtils;
 import android.view.MenuItem;
 import android.view.View;
+import android.widget.EditText;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.TextView;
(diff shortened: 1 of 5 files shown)


JamesSmartCell:
@ChintanRathod this can now be completed, can you move your branch to 'Component Upgrade'. I have added an 'ENS Name' control to the screen, which is used for locating an ENS name for users who haven't set their Reverse Resolve contract - it works the other way - given an ENS name, if it resolves to the current wallet address then the ENS name is set.

If there is no ENS name, and the user types a wallet name, then this could be assigned to be the wallet name.

If the user overrides their ENS name with a non-ens wallet name, then we'd be using that. So, when the user leaves the activity just assign whatever is in the input box to the name. However, I don't think we should allow the user to set anything that matches a call to ```ENSHandler.canBeENSName(...)``` unless it resolves to the wallet. If this is tried then it should show an error.

JamesSmartCell:
This one might have to be refactored ... 

Use the 'ENS Name' input view; once something is typed in there, complete the check to see if it's actually an ENS name or not.

If it's not, use wallet.name, otherwise keep the code that sets it as wallet.ENSname.

When displaying; first show wallet.ENSname, if there's no wallet.ENSname then show the wallet.name

Can you add a method into ```class Wallet``` to return the name to do this, and use this method where wallet.ENSname or wallet.name is required.

JamesSmartCell:
See this issue:

https://github.com/AlphaWallet/alpha-wallet-android/issues/1795

This new issue superseeds the info above, except add the 'Scan ENS' text so user can put their ENS name into the wallet name.

Angie3120 (approved):

Angie3120 (approved):

mpaschenko:
@JamesSmartCell applied code review fixes

PR #1679, opened 2020-11-30
Fix for tokens not getting updated correctly.

### LINKED ISSUE #1675: Incorrect Balance for some tokens
Seems like a caching issue, as the users are reporting incorrect balances after they have received tokens. Mainly on xDAI network but also have instances where it was reported for mainnet.

The other tokens reported are:
Aave on main-net
[HNY](https://blockscout.com/poa/xdai/tokens/0x71850b7E9Ee3f13Ab46d67167341E4bDc905Eef9/token-transfers) on xDAI network
[LINK](https://blockscout.com/poa/xdai/address/0xE2e73A1c69ecF83F464EFCE6A5be353a37cA09b2/transactions) on xDAI network
TRISKEL on Gorli (Test) network 

Candidate for a hotfix as we have a strong user base on xDAI.

JamesSmartCell:
Fixes #1675 

Fixed with a small refactor in TokensService:

Remove the separate list of tokens in the TokensService, which controlled updates. This list could get out of sync with the app's known tokens in the database.

The separate list was replaced with a call to fetch all the active tokens from the database; we then find the token with the highest update priority (taking into account if the token has value; is it on main chain or a test chain, when was it last updated etc). This token is then marked as updated so it won't get picked again and the app goes on to attempt a balance update.

Now the TokensService update operates entirely from the database; this way we don't have to keep a separate list in sync with a master list; less maintenance, less to go wrong.


James-Sangalli (changes requested):
It's working in that it addresses the issue but another issue is present (it show's as pending when it is confirmed since a while ago)

Also, why is google chrome opened instead of the dapp browser when clicking "view on etherscan"?

actual
pending

JamesSmartCell:
Maybe raise a new issue with (if possible) a repro case.
Opening etherscan in the phone's default browser is the current behaviour.

Neither of these points are related to this PR.

James-Sangalli (approved):
@JamesSmartCell made a new issue for the pending thing.

JamesSmartCell:
Also fixes #1687

PR #1710, opened 2021-01-25
added Eporio dapp

alexsyo:


JamesSmartCell:
@alexsyo I will ask our web front end to add this as well - the list maintained here is for when a user types something into URL for auto-complete, eg typing '1in...' will open a well known token swapping site.

Issue #1596, opened 2020-09-11
Replace Network name with Home button in the browser


colourfreak:
I think it's more useful to have a Home button (linked to alphawallet.com/browser/) instead of Ethereum huge button. See the top right corner next to the address input field and menu icon. 

![Browser_Device](https://user-images.githubusercontent.com/51817359/92924372-c0e40d00-f430-11ea-91eb-c139f4f24d69.png)

Icon:
https://app.zeplin.io/styleguide/5e0b15e46049c216fb7ab4d5/components?coid=5f5b6aa6deccf16974676595

Zeplin:
https://zpl.io/2Ze0ppJ

Selecting a Network and go into the three dots menu.

colourfreak:
Any update on this one? It will be very handy.

colourfreak:
please remember about displaying Network name in the top left https://zpl.io/Vq7LmA5

colourfreak:
https://zpl.io/Vq7LmA5

<img width="1748" alt="Screenshot 2021-04-21 at 12 55 15" src="https://user-images.githubusercontent.com/51817359/115542794-ee374a00-a2a0-11eb-986a-d3f27ed4a03f.png">
<img width="1722" alt="Screenshot 2021-04-21 at 12 55 48" src="https://user-images.githubusercontent.com/51817359/115542807-f1cad100-a2a0-11eb-93d1-36ffbe28bc64.png">

colourfreak:
Let me know if you have questions @mpaschenko

PR #1715, opened 2021-02-05
fix for browser url bar resizing

### LINKED ISSUE #1516: Back button in the browser doesn't go to the previous web page
Android's back button in the browser doesn't go to the previous web page Instead, it goes to the previous tab. The user has to hit the "Back" button on the upper-left-hand in order to go back. This issue asks it to function the same

You might ask: If we allow the back button behaves like a normal browser, then how does the user go back to the token card the user was at?

The answer is the user can't go to the browser tab unless he leaves the token card he was at first, so there is no token card to go back to. Therefore it should be okay to let back button function as browser back button when in the browser.

The second issue is that the back button will never bring the user back to other tabs if the current user has a long browsing history.

That is true but the user can always switch to other tabs with the bottom row of icons. In the long run, I prefer to think the tabs we have "Wallet, Transactions, Settings" are like browser tab (I know we don't have tabbed browsing yet but we will get there), so back is always done within the current tab.

Sending it to @tomekalphawallet first so if he approves my request, this goes to @ChintanRathod with @AW-STJ in the loop.

JamesSmartCell:
We were seeing an issue where even after user dismisses the soft keyboard, the URL bar wouldn't collapse back into the default state.

This mod keeps the URL bar expand (where the URL bar is expanded to fill the whole screen width while user typing in it, similar to most browser apps) hooks to where the URL bar gains focus which is 100% reliable.

The URL bar collapse is now hooked to the softkeyboard being dismissed, which is the desired behaviour. Previously it was hooked in different places to where we'd expect the user is doing a different activity.

Closes #1516

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
+++ b/app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
@@ -133,6 +133,7 @@ import static com.alphawallet.app.entity.CryptoFunctions.sigFromByteArray;
 import static com.alphawallet.app.entity.Operation.SIGN_DATA;
 import static com.alphawallet.app.entity.tokens.Token.TOKEN_BALANCE_PRECISION;
 import static com.alphawallet.app.ui.MyAddressActivity.KEY_ADDRESS;
+import static com.alphawallet.app.util.KeyboardUtils.showKeyboard;
 import static com.alphawallet.app.widget.AWalletAlertDialog.ERROR;
 
 public class DappBrowserFragment extends Fragment implements OnSignTransactionListener, OnSignPersonalMessageListener, OnSignTypedMessageListener, OnSignMessageListener,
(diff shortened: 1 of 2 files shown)


colourfreak:
1. Add a Home Button as I requested in #1596 
2. To collapse the keyboard and escape from typing, you can click anywhere in the pink area.
3. Add "Set as Home Page" and "Change Network" to the menu

Screenshot 2021-02-05 at 09 58 02

JamesSmartCell:
@colourfreak Yes, I'll be upgrading the UI to your spec but it's outside the scope of this PR.

Also what about network display? I have found that it's very useful to have the active network being displayed, and also the ability to quickly change networks. Can we at least have a small round token next to the home icon to display the current network as a colour similar to MetaMask?

colourfreak:
@JamesSmartCell Instead of "Change Network" we can say "Network: Ethereum" or "Network: Ropsten". When you tap on this item, you will see a menu where you can choose a network as it is right now. 
Screenshot 2021-02-10 at 08 08 12

JamesSmartCell:
Ok, yes I can add that but I'm talking about a visual cue. What about a token behind the 'home' logo?

Have a look at MetaMask, it's done very well there with the small round colour token that shows network at a glance.

PR #1716, opened 2021-02-06
Fix for transaction speedup and cancel

### LINKED ISSUE #1713: Speed-up transaction either crashes app or the flow goes into loop
The speed-up transaction does not work as expected. 

The app crashed the first time I tried to open up speed up transactions.

The flow is also weird now. The speed up transaction goes on like this:
1. Click pending transaction
2. Click details button on top right
3. Click 'Speedup Txn' button
4. Cick 'settings' button at top-right corner (Screen is has a dialogue box saying 'Submit replacement txn')
    Go backs back to Step 2.
- OR -
5. Click the blue dialogue box and app crashes

image

Video:
https://drive.google.com/file/d/1S9oirBr_APSRD1kwHcVDhCgE2LXRC3KT/view?usp=drivesdk

JamesSmartCell:
- Fixes crashbug in speedup transaction. Closes #1713 
- Adds cancel transaction back in, fully tested with DeFi transactions on mainnet.
- Fix gas price calculation for re-submitting transaction (needs to be at least 10% higher than transaction it's replacing).

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -182,6 +182,10 @@
             android:label="@string/title_send_settings" />
 
         <activity
+            android:name=".ui.GasSettingsActivityLegacy"
+            android:label="@string/title_send_settings" />
+
+        <activity
             android:name=".ui.SellDetailActivity"
             android:label="@string/action_sell"
             android:windowSoftInputMode="adjustPan" />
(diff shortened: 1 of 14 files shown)


AW-STJ:
Not sure whether it is the problem in how I made the build. But the application only shows 0 balance for all my tokens. And crashes when I try to send transactions.

Video: https://drive.google.com/file/d/1UHg80pSqn6FHLL_qjU6AhnUBVYnm7n39/view?usp=drivesdk

PR #1731, opened 2021-02-20
Fix for custom nonce not being used

### LINKED ISSUE #1727: Allow users to set Nonce
The current nonce in Custom Gas is not working.

image

JamesSmartCell:
Closes #1727 

Nonce was not being passed to transaction push.


colourful-land (approved):
Looks good & let's avoid using "trans" for transactions. Many our users might think trans-gender

JamesSmartCell:
> Looks good & let's avoid using "trans" for transactions. Many our users might think trans-gender

'Speedup Transaction' looks terrible on the UI. There's no enough space in the button for the text. Originally I put 'Speedup TX' but that is not ideal either.

JamesSmartCell:
> Looks good & let's avoid using "trans" for transactions. Many our users might think trans-gender

'Speedup Transaction' looks terrible on the UI. There's no enough space in the button for the text. Originally I put 'Speedup TX' but that's not ideal.

PR #1732, opened 2021-02-20
Add xDai bridge

### LINKED ISSUE #1707: Hardcode 'Convert xDAI' button to https://bridge.xdaichain.com/
Refer to https://github.com/AlphaWallet/alpha-wallet-ios/issues/2473

Like what we do for swaps (https://1inch.exchange)

JamesSmartCell:
Add ability to make custom bridges between tokens and dapps (should be done with TokenScript!)

Should also add the xDai -> DAI bridge too.

It touches a lot of files, but unfortunately that's needed as the best way to get from a leaf activity, back to a specific root fragment is to go via a return intent; which involved changing the Context to Activity in a few places. 

Closes #1707


colourful-land:
It's hard to make the correct TokenScript for this anyway because this involves the anticipation that some money will arrive in a different chain.

JamesSmartCell:
It's a very simple shim.

An action button 'xDai Bridge' is added to the DAI and SAI token detail page (that appears after you click on a token in the wallet).
On clicking this button, the app opens the dappbrowser at the xdai bridge app.

Having a simple script to do this that dapp authors can submit may a nice hook to get them started. What do you think?

colourful-land (approved):
Will later find a way to include such static "dapp" into TokenScript nicely.

PR #1758, opened 2021-03-06
Empty Transactions view

### LINKED ISSUE #1744: Add a custom graphic for empty Activities
https://zpl.io/aX6D9yl

Screenshot 2021-02-24 at 11 12 43

JamesSmartCell:
Closes #1744 

Can you check the Translations?


JamesSmartCell:
> I don't know why you changed the LinearLayout to Relative Layout in app/src/main/res/layout/fragment_transactions.xml - approved anyway with Chinese translation update.

@colourful-land That was so the placement of the new 'empty transactions' graphic would be 100% deterministic, rather than where the OS feels like it should go depending on existing views. Previously it didn't really matter because the empty tx view was quite small and we didn't really care; these days I get scolded by Tomek if the layout isn't right.

PR #1764, opened 2021-03-09
Fix for issue where xdai transfers weren't picked up

JamesSmartCell:
xDai and other blockscout powered chains didn't detect ERC20 & NFT token transfers due to a bug in the API fetch.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/repository/EthereumNetworkBase.java
+++ b/app/src/main/java/com/alphawallet/app/repository/EthereumNetworkBase.java
@@ -63,7 +63,7 @@ public abstract class EthereumNetworkBase implements EthereumNetworkRepositoryTy
     public static final String BACKUP_INFURA_KEY = BuildConfig.XInfuraAPI;
     public static final String MAINNET_FALLBACK_RPC_URL = !getAmberDataKey().startsWith("obtain") ? "https://rpc.web3api.io?x-api-key=" + getAmberDataKey() : MAINNET_RPC_URL;
     public static final String CLASSIC_RPC_URL = "https://www.ethercluster.com/etc";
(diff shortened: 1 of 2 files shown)


colourful-land (approved):
Approved, although I didn't get why checkTransaction becomes writeTransaction

JamesSmartCell:
> Approved, although I didn't get why checkTransaction becomes writeTransaction

Because where previously we were checking to see if we'd recorded the transaction, and if not fetch it, to simply forming the transaction and writing it directly based on the information we have. (Previously I was cautious to see if this system would work, now it works we have enough info to make a shortcut).

sharonsylu:
Sorry I could not find the issue ticket which this PR corresponds to, so putting this issue here (@JamesSmartCell please tell me if I should delete and move elsewhere)

Observed: 

Historical xDai transactions now populating and displaying in Activity, but not picking up latest activity for over past 7 days (RealT Tokens on xDAI POAP). 

Screenshot of App showing last txn picked up was 14 June: 
photo_2021-06-28_19-08-36 (3)


Screenshot of Blockscout showing 2 transactions since 14 Jun: 
Screen Shot 2021-06-29 at 4 13 40 pm


Actions taken: 
Have asked user to check they have updated  app v3.33
Have asked user to Navigate to Settings --> Advanced --> Clear Browser Cache. 


Waiting to hear back to see if this resolves, and will update this to confirm.

If no resolution, then possibilities: 
- asychronous delays in how Alphawallet fetches/calls txn history from APIs (Etherscan/Blockchain) 
- asychronous delays in how Blockscout fetches transactions and feeds into API 
- not a time delay, but something else that changed that stops Alphawallet from picking up token transfer activities eg, custom routes ?

PR #1765, opened 2021-03-09
Fixing issue where network couldn't be picked in dappbrowser

JamesSmartCell:
Reproducible:

1. Enable a lot of networks in settings.
2. Pick one of those (not ethereum main net) in the dapp browser
3. Go back to settings and disable all the networks except main net
4. Go to dapp browser and see that no network is selected.

Fix: If the network was deselected and is no longer in the enabled filters then revert network to main net.


colourful-land:
I don't agree with the fix, because the software made a decision that the user isn't aware off - imagine the user making an angry call to the blockchain's operator only to find a few hours later that αW swapped to default.  Since this doesn't often happen, a modal is acceptable and the user might be presented with a choice that he can't skip. @colourfreak ? Also @AW-STJ @hboon  how does iOS handle such situation? Might need some align-up here.

colourful-land (changes requested):
I don't agree with the fix, because the software made a decision that the user isn't aware off - imagine the user making an angry call to the blockchain's operator only to find a few hours later that αW swapped to default.  Since this doesn't often happen and is crucial for the dapp to work, a modal is acceptable and the user might be presented with a choice that he can't skip. Recall that I likened a modal dialogue as a "but, sir!" in human conversation and this is the time when a servant should have said "but, sir!" @colourfreak ? Also @hboon  how does iOS handle such situation? Might need some align-up here.

P.S. this issue is pretty trivia might the intern / outsource developer do it? @AW-STJ

JamesSmartCell:
1. User chooses a network for dappbrowser.
2. Subsequently, user disables that network in the wallet.
3. Dappbrowser can't use that network as it's been told not to use that network.

With your solution, dappbrowser continues to use the network after the user has instructed not to use that network.

Imagine this scenario: 

user is using an xDai dapp, but wants to ensure, for whatever reason that the wallet won't use xDai, so they switch that network off.

Dappbrowser was previously set to xDai, and unless we do something will still be switched to xDai.

This is quite clearly NOT what was intended as xDai network was switched off.

hboon:
> @hboon how does iOS handle such situation? Might need some align-up here.

iOS falls back to mainnet (actually this is wrong too because the behaviour is unpredictable if mainnet was disabled).

I *think* there isn't a risk in losing funds from sending the transaction to the wrong network since the chain ID is embedded, and most apps should/ought to check for it before even accepting the user. Maybe loss of gas?

So we have a few options here.

colourful-land:
> > @hboon how does iOS handle such situation? Might need some align-up here.
> 
> iOS falls back to mainnet (actually this is wrong too because the behaviour is unpredictable if mainnet was disabled).
> 
> I _think_ there isn't a risk in losing funds from sending the transaction to the wrong network since the chain ID is embedded, and most apps should/ought to check for it before even accepting the user. Maybe loss of gas?

1. User enters testnet and she sees her balance is Ξ300
2. User disables the testnet from the available networks
3. The user is taken to mainnet without telling the user, and the system shows a balance of Ξ250
4. The user sends Ξ100 to bob thinking she sent testnet Ether. Bob receives Ξ100 real ether.

You can see cryptography doesn't come to play in such a scenario. So why chain ID comes to this discussion?

hboon:
> 4\. The user sends Ξ100 to bob thinking she sent testnet Ether. Bob receives Ξ100 real ether.
> You can see cryptography doesn't come to play and this is purely UX problem. So why chain ID comes to this discussion?

Oh right. Serves me right for multi-tasking.

I suppose the only way that makes sense is to prompt the user (by showing the selector) until a selection has been made then.

colourful-land (approved):
Got blamed for not approving. Approved. I didn't agree with this fix but I didn't have time to do it properly and don't want to be seen as not approving. Maybe I start a new thread on setting up a modal dialogue and accept this as half-fix for now.

JamesSmartCell:
Accept that this solution isn't ideal - since it's very much an edge condition; it's probably acceptable to pop a 'choose new network for DappBrowser' dialog.

@colourfreak What do you think?

colourfreak:
What could solve a problem is 
- adding network name to all action sheets https://zpl.io/VO4dq4j
- add info in the top left corner about what network you are connected to https://zpl.io/Vq7LmA5

And if you do steps like James mentioned, simply don't change anything.

colourfreak:
Screenshot 2021-03-15 at 13 23 07

Scenario:
1. You are connected to xDAI
2. You go to Network Selection
3. You go to Select Active Networks and deselect 'xDai'
4. Since you were connected to xDAI, you will see a dialog.
5. If you choose: 'keep connected' it does not deselect xdai and keep you connected to xDAI
6. If you choose: 'disconnect' it deselect xdai from Active Networks and changes to Ethereum Mainnet automatically.

JamesSmartCell:
> Screenshot 2021-03-15 at 13 23 07
> 
> Scenario:
> 
> 1. You are connected to xDAI
> 2. You go to Network Selection
> 3. You go to Select Active Networks and deselect 'xDai'
> 4. Since you were connected to xDAI, you will see a dialog.
> 5. If you choose: 'keep connected' it does not deselect xdai and keep you connected to xDAI
> 6. If you choose: 'disconnect' it deselect xdai from Active Networks and changes to Ethereum Mainnet automatically.

Yes, putting the dialog there seems the most logical place you'd expect to see it.

colourfreak:
https://zpl.io/a7mv7zR
https://zpl.io/Vq7LmA5

PR #1766, opened 2021-03-10
Add Heco chain

### LINKED ISSUE #1702: Supporting Heco
Integration Docs Here:

https://docs.hecochain.com/#/en-us/wallet

Thanks a lot.

[Heco icon.zip](https://github.com/AlphaWallet/alpha-wallet-android/files/5742914/Heco.icon.zip)

JamesSmartCell:
- Closes #1702 
- Use transaction time to determine ticker validity


colourful-land (approved):
After it has been done the 5th time, we should automata the process into a config, so if someone provides an SVG it's put into the source code and the VectorDrawable is produced at compile time, therefore making the change leads to iOS and Android supporting the same set of networks at the same time.

JamesSmartCell:
Yes, I'm going to work on a refactor for the node connection. It will take into account what Web3j have been working on with their 'universal connection' to eg Besu, Corda, Quorum, and also the 'ethereum-chains' in the ethereum github repo to allow dapps to add a network, and users to add a network chosen from that list.

PR #1776, opened 2021-03-15
Ensure walletconnect works everywhere

### LINKED ISSUE #1719: Make scan QR code button in browser function as Universal scanner
The current QR code really only works only to point the browser to other websites. The payment request and WalletConnect  QR codes are not supported.

JamesSmartCell:
Fixes #1719


colourful-land (approved):
I only tested this against walletconnect

PR #1781, opened 2021-03-17
Fix for Euro Input

JamesSmartCell:
Fixes an issue with entering European mainland style input (eg 100,00).

Also refactor and optimise the ENS check (issue seen in the posted video): No need to check for ENS if input cannot be ENS.

https://user-images.githubusercontent.com/12689544/111546455-d2391800-87cb-11eb-8e07-5c2bc0e4add9.mp4

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/widget/entity/ENSHandler.java
+++ b/app/src/main/java/com/alphawallet/app/ui/widget/entity/ENSHandler.java
@@ -34,6 +34,7 @@ import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.schedulers.Schedulers;
 
+import static com.alphawallet.app.util.AWEnsResolver.couldBeENS;
 import static org.web3j.crypto.WalletUtils.isValidAddress;
 
 /**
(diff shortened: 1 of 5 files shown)


colourful-land (changes requested):
What's the reason to use home-grown parser instead of [DecimalFormat.parse()](https://developer.android.com/reference/java/text/DecimalFormat)?

colourful-land (changes requested):
Please don't convert Number into double! You risk getting 0.9999999 out of 1

JamesSmartCell:
Then can you just re-code the routine to how you think it should be and I can take it from there? I checked StackOverflow for what people do for actual production code and my original routine is closest to what is the go.

colourful-land:
> Then can you just re-code the routine to how you think it should be and I can take it from there? I checked StackOverflow for what people do for actual production code and my original routine is closest to what is the go.

Happy to do that but can you correct this:

> The default Locale is constructed statically at runtime for your application process from the system property settings, so it will represent the Locale selected on that device when the application was launched. Typically, this is fine, but it does mean that if the user changes their Locale in settings (system or in-app) after your application process is running, the value of getDefaultLocale() probably will not be immediately updated.

colourful-land (approved):
looks good

PR #1786, opened 2021-03-19
Switch to backup node if rate limited.

JamesSmartCell:
Preparing for deploying and using our backup nodes.

If we are rate limited or the node becomes unresponsive then switch to the backup node.


colourful-land (approved):
This is hard to test so I'll just approve since the code look okay

Issue #1775, opened 2021-03-15
Allow camera access for Ramp integration


AW-STJ:
User reported issue. I could not reproduce this error, as Ramp does not prompt me to access the camera for KYC in Australia.

![image](https://user-images.githubusercontent.com/61957841/111102203-a0dc0480-859f-11eb-9722-da59ce3f0c1a.png)

Can this be fixed without having the steps to reproduce. I suspect the user is from Mali, based on Mixpanel insights.

PR #1787, opened 2021-03-21
Allow request camera permission for dappbrowser

justindg:
Potential fix for #1775

I cannot replicate the exact issue for Ramp, but what I noticed is that the Dapp Browser is auto-denying requests to use the camera on other websites.

This PR addresses this issue by displaying a simple dialog that prompts the user to Allow or Deny camera access on websites that require one.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
+++ b/app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
@@ -5,10 +5,12 @@ import android.animation.Animator;
 import android.animation.LayoutTransition;
 import android.animation.ValueAnimator;
 import android.app.Activity;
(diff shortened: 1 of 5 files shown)


JamesSmartCell (commented):
Is this the recommended way to obtain permission for things now? As opposed to the old default 
```
final String[] permissions = new String[]{Manifest.permission.CAMERA};
ActivityCompat.requestPermissions(this, permissions, RC_HANDLE_CAMERA_PERM); 
```

JamesSmartCell:
Hi @justindg can you check the comment? Is there a reason to use custom code in this place instead of using the Android libraries?

JamesSmartCell (approved):
It works after a patch. Didn't work as it stood.

PR #1794, opened 2021-03-27
Action sheet updates

jinkjonks:
Screenshot_20210327_230314

Created BalanceDisplay widget and AmountDisplay widget.

Also added Token Icon and ChainName.
ActionSheetDialog now looks as pictured.


app/src/main/java/com/alphawallet/app/widget/AmountDisplay.java
```diff
@@ -0,0 +1,46 @@
...
+import java.math.BigInteger;
+import java.text.NumberFormat;
+import java.util.Locale;
+
+import javax.annotation.Nullable;
+
+/**
+ * Created by Jenny Jingjing Li on 21/03/2021
+ * */
+
```
JamesSmartCell, line 20: This should be called 'AmountDisplayWidget' for consistency with the other widget classes

app/src/main/java/com/alphawallet/app/widget/BalanceDisplay.java
```diff
@@ -0,0 +1,80 @@
...
+
+        chainIcon = findViewById(R.id.chain_icon);
+
+    }
+
+    public void setupBalance(Token t, TokensService ts ,Activity act)
+    {
+        token = t;
+        activity = act;
+        tokenService = ts;
```
JamesSmartCell, line 54: You're only using TokensService locally so no need to make it a member of the class.

app/src/main/java/com/alphawallet/app/widget/BalanceDisplay.java
```diff
@@ -0,0 +1,80 @@
...
+
+        chainName.setChainID(token.tokenInfo.chainId);
+        chainName.invertChainID(token.tokenInfo.chainId);
+        chainName.setVisibility(View.VISIBLE);
+        chainIcon.bindData(tokenService.getToken(token.tokenInfo.chainId, token.getAddress()), null);
+
```
JamesSmartCell, line 60: To get the base chain, use method ```getCurrentAddress``` from TokensService. This will put the network/chain icon into the icon view. You can also use "eth" for the address like this:

```getToken(chainId, "eth")```

To see the issue, try sending an ERC20 token with the source code as it stands; you'll see the icon doesn't appear as the chain icon.

File: app/src/main/java/com/alphawallet/app/widget/BalanceDisplay.java [hunk 4]
```diff
@@ -0,0 +1,80 @@
...
+import com.alphawallet.app.util.BalanceUtils;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+
+/**
+ * Created by Jenny Jingjing Li on 4/3/2021
+ */
+
+public class BalanceDisplay extends LinearLayout
```
JamesSmartCell, line 24: This should be BalanceDisplayWidget for consistency.

JamesSmartCell (changes requested):
@jinkjonks nice, this is very close to how it should be.

app/src/main/java/com/alphawallet/app/widget/BalanceDisplayWidget.java
```diff
@@ -0,0 +1,77 @@
...
+
+        token = null;
+
+        chainIcon = findViewById(R.id.chain_icon);
+
+    }
+
+    public void setupBalance(Token t, TokensService tokenService, Activity act)
+    {
+        token = t;
```
JamesSmartCell, line 50: token and activity don't need to be members, and you don't need Activity as the widget has Context to retrieve the resource.

JamesSmartCell: Actually Token does need to be a member, I didn't see the usage in the final function.

app/src/main/java/com/alphawallet/app/widget/BalanceDisplayWidget.java
```diff
@@ -28,50 +26,39 @@
...
         chainIcon.bindData(tokenService.getToken(token.tokenInfo.chainId, tokenService.getCurrentAddress()), null);
 
-        balance.setText(activity.getString(R.string.total_cost, token.getStringBalance(), token.getSymbol()));
+        balance.setText(getContext().getString(R.string.total_cost, token.getStringBalance(), token.getSymbol()));
     }
 
-    public void setNewBalanceText(BigDecimal transactionAmount, BigInteger networkFee, BigInteger balanceAfterTransaction, boolean isSendingTransaction)
+    public void setNewBalanceText(Token token, BigDecimal transactionAmount, BigInteger networkFee, BigInteger balanceAfterTransaction, boolean isSendingTransaction)
     {
-        balance.setText(activity.getString(R.string.total_cost, token.getStringBalance(), token.getSymbol()));
```
JamesSmartCell, line 50: Just two small changes to this function:

- The formatting is a bit off now; there's spaces and inconsistent brace formatting.
- Now you've increased the clarity of the function it can be seen that 'isSendingTransaction' is actually not required, because it'll always be true if this function is getting called and if token isn't ethereum.

Note that this widget will get more complex when we add NFT handling, but that isn't happening yet.

JamesSmartCell (approved):

PR #1810, opened 2021-04-17
Add logic to separate mainnet and testnet coins

### LINKED ISSUE #1804: Make 'Testnet' more obvious and separated from 'Mainnet'
Clickable prototype (iOS):
https://invis.io/GU10Q3YYSZB7

- You have a toggle to enable Testnets. This toggle disables Mainnet and enable Testnet. And other way around. Only one can be enabled at the time.
- After you hit the toggle, it warns you about what testnets are
- We say that they have no value at all time.
- When you are viewing a Wallet in a Testnet, all Mainnet tokens are hidden.
- You have a special link that on tap shows a modal explaining "Where are my Tokens?"
- You can go to "Switch to Mainnet" which will go to Select Active Networks again

Zeplin:
https://zpl.io/VDgG0Dg

testnet_flow_android

ChintanRathod:
Fixes #1804


ChintanRathod:
@JamesSmartCell This PR has no following changes as has new design. 

image

app/src/main/java/com/alphawallet/app/repository/SharedPreferenceRepository.java
```diff
@@ -23,6 +23,7 @@
     public static final String USER_LOCALE_PREF = "user_locale_pref";
     public static final String HIDE_ZERO_BALANCE_TOKENS = "hide_zero_balance_tokens";
     public static final String FULL_SCREEN_STATE = "full_screen";
```
JamesSmartCell, line 25: Typo

app/src/main/res/values/strings.xml
```diff
@@ -762,4 +762,8 @@
     <string name="action_allow_webkit_camera">Allow</string>
     <string name="action_deny_webkit_camera">Deny</string>
     <string name="message_permission_request_webkit_camera">%s wants to use your camera</string>
+    <string name="mainnett">Mainnet</string>
```
JamesSmartCell, line 765: typo (mainnet)

JamesSmartCell (changes requested):
Looks good! Can you just fix the two typos and I'll push it.

app/src/main/java/com/alphawallet/app/repository/SharedPreferenceRepository.java
ChintanRathod: Done

app/src/main/res/values/strings.xml
ChintanRathod: Done

JamesSmartCell (changes requested):
Just one issue I noticed: When starting the wallet for the first time with this patch, it was still showing all the testnets and mainnets.

Can you add a pass-over the selected networks at runtime in HomeViewModel to ensure that either only testnet or mainnet is selected depending on the switch setting.

ChintanRathod:
I guess I know the problem for fresh installation. Looking into it.

ChintanRathod:
@JamesSmartCell When I do fresh installation, I got this screen. Could you confirm this is the same you got?

image

Because this is perfect implementation as expected. I found no issue with the patch.

ChintanRathod:
I accidently changed the layout and didn't notice the use of the same layout in other activities. So created separate layout file for the SelectNetwork activity and revert the activity_list changes.

JamesSmartCell:
Hi @ChintanRathod no the opposite, it's for existing installation. Say you have a lot of networks selected, you upgrade to a version with this feature, but still all testnet and mainnet tokens are together until you go to the 'selectnetwork' screen.

ChintanRathod:
Ok. Let me try to install older app and try with this new one.

hboon:
BTW, don't forget to check the dapp browser (and WalletConnect connection screen) in case those are affected.

JamesSmartCell:
> BTW, don't forget to check the dapp browser (and WalletConnect connection screen) in case those are affected.

Yes I checked that, that code isn't touched so it should still work as before, but I'll check before pushing

hboon:
> > BTW, don't forget to check the dapp browser (and WalletConnect connection screen) in case those are affected.
> 
> Yes I checked that, that code isn't touched so it should still work as before, but I'll check before pushing

Ok. But just in case: I mean like if you enable mainnets in Settings, dapp browser shouldn't let you switch to a testnet. Similarly for WalletConnect.

JamesSmartCell:
> > > BTW, don't forget to check the dapp browser (and WalletConnect connection screen) in case those are affected.
> > 
> > 
> > Yes I checked that, that code isn't touched so it should still work as before, but I'll check before pushing
> 
> Ok. But just in case: I mean like if you enable mainnets in Settings, dapp browser shouldn't let you switch to a testnet. Similarly for WalletConnect.

It won't let you do that; you can't switch to a network that's not active.

When the PR is complete I'll test disabling a network that is currently active too.

JamesSmartCell (approved):

PR #1834, opened 2021-05-13
add French to locales in class LocaleRepository

desouzag:
Added French to locales in class LocaleRepository


JamesSmartCell:
Hi @desouzag, thankyou for committing this PR :) The Travis build error is indicating there's unescaped apostrophes, so eg in this line:

```<string name="body_low_gas">Gas price is set below the 'slow\' speed.\nYour transaction may take a long time to be written or may never be written.</string>```

It needs to be like this:

```<string name="body_low_gas">Gas price is set below the \'slow\' speed.\nYour transaction may take a long time to be written or may never be written.</string>```

Likewise, anywhere where the ```'``` character is used in the strings.xml file, it must be escaped with the ```\``` character.

desouzag:
Hello @JamesSmartCell , thank you. 

I'll update it and submit again.

JamesSmartCell:
Hi @desouzag do you have any update on the translations yet?

desouzag:
Hi @JamesSmartCell, I was sick all week. I will be able to publish an update by the end of next week. I am almost finish, 200 lines more or less.

PR #1837, opened 2021-05-17
ERC721 Tokens Transaction Sheet

jinkjonks:
Sending ERC721 Tokens on the app is now done via the ActionSheetDialog. 
 

An AssetDetailView was also implemented so that users can see what ERC721 Token they're sending before they confirm the transaction.


JamesSmartCell (approved):
Nice work, congrats on your first big PR in the repo.

PR #1862, opened 2021-05-24
strings.xml translation

desouzag:
File completely translated.
From the most recent update.


JamesSmartCell:
Nice work @desouzag! To finish you only need to add the French locale to the locale list:

```class LocaleRepository```
...
```
    private static final String[] LOCALES = {
            "en",
            "zh",
            "es",
            "fr"   //<---
    };
```

Then you will be able to see the new 'French' setting. If your phone is already set to French locale it should pick this up automatically. If not, I can fix it.

PR #1899, opened 2021-06-18
SignatureV4 Eth node access sample code

samuelhbne:
This draft branch was in purpose of demonstrate AWS Etheurem node access with Signature V4 API. It SHOULD NOT be merge into master branch directly. An "ETH NODE ACCESS" entry has been added into support setting page, which triggers the AWS Etheurem node access sample code. The returned response will be write into log.


Memohalliwell (commented):
> 

>

Issue #1868, opened 2021-05-26
Add Custom Linking to Quickswap for tokens on Polygon


colourfreak:
#2103 - since Quickswap is a for of Uniswap, we can add custom linking for all tokens on Polygon.

![Quickswap_AlphaWallet](https://user-images.githubusercontent.com/51817359/119621863-e79f7380-be06-11eb-8a17-7dd34c0ff951.jpg)

colourfreak:
Third and last task @mpaschenko @AW-STJ

mpaschenko:
@JamesSmartCell @colourfreak just checked the code, but can't find any implementation for that (I mean for the other tokens). But it exists on ios, should I make it from scratch? Or did I miss something?
Also for the iOS for the LINK tokens it redirects to **https://1inch.exchange/#/r/0x98f21584006c79871F176F8D474958a69e04595B/LINK/ETH** should I use quickswap or 1inch as on iOS?

JamesSmartCell:
We have previously added custom functions like this. It should be fairly trivial to extend this, for reference check ```openDapp(C.XDAI_BRIDGE_DAPP);``` in Erc20DetailActivity.

maybe 1inch on Mainnet, quickswap on Polygon?

hboon:
1inch also supports Polygon now. So if it’s good enough, we can avoid QuickSwap entirely (less code to add an maintain). @colourfreak do you know?

colourfreak:
Polygon users use Quickswap a lot. Much more than 1inch.

PR #1962, opened 2021-08-06
Added custom linking to quickswap (#1868)

mpaschenko:


app/src/main/java/com/alphawallet/app/ui/Erc20DetailActivity.java
```diff
@@ -328,6 +328,9 @@ public void handleClick(String action, int actionId)
         if (actionId == R.string.convert_to_xdai)
         {
             openDapp(C.XDAI_BRIDGE_DAPP);
+        } else if (actionId == R.string.swap_with_quickswap) {
+            String queryPath = "?use=v2&inputCurrency=" + token.getAddress();
```
hboon, line 332: Remember to use "ETH" (verbatim) instead of `0x00...00` if it is the native crypto token according to Uniswap docs https://uniswap.org/docs/v2/interface-integration/custom-interface-linking/

JamesSmartCell (changes requested):
@mpaschenko please see @hboon's comment for this line:

```String queryPath = "?use=v2&inputCurrency=" + token.getAddress();```

Suggest you can use ```token.isEthereum() ? ETH_SYMBOL : token.getAddress();``` instead of ```token.getAddress()```

mpaschenko:
updated as requested

JamesSmartCell (approved):

JamesSmartCell (approved):

PR #1964, opened 2021-08-10
1887 add custom rpc network

mpaschenko:


JamesSmartCell (changes requested):
Can you update the translation strings?

JamesSmartCell (changes requested):
- hook up the IME 'Next' on the soft keyboard.
- Chain ID should be numerical input
- Network Name should capitalise first letter of each word
- Symbol should be capitalised by default
- URL input fields should be 'URL' type with the forward slash on the soft keyboard.
  
And for operation:

- 'Block Explorer URL' should be 'Transaction Explorer URL' and give a 'ghosted' example hint which could be: ```https://etherscan.com/tx/```
- Add in a new field which is 'Explorer API' which is an interface for scanning transactions and usually ends with ```/api?```. The hint could be ```https://api.etherscan.io/api?```

The reason people need to add these on the end is due to inconsistent naming by some API developers. Most use the same format with '/api?' on the end. This may benefit from a little refactoring as it's only covalent that doesn't follow this format (I can add this in after you have updated the PR).

This makes it a lot easier for users to see what they're supposed to be typing into these fields. It's not really clear, even on metamask.

It seems to work well though - I tested it on a network that I added.

JamesSmartCell:
One more thing - there needs to be a way users can edit/update the settings, even for supplied networks.

Initially, I would suggest you use the 'long press' method in the 'Select Active Networks' Activity. This would just take you to your edit page, with the relevant fields populated. Once done, we can ask Tomek for a better solution. Maybe putting an 'edit' menu button next to the (+) menu button in the top right?

I'd also like for users to be able to set a URL for the chain logo, but that is another PR.

mpaschenko:
@JamesSmartCell as you wrote above about editing of the supplied networks - there are a number of the related hardcoded constants colors/icons/chain_id etc. in the code - it's quite a big task to add editing for that values. 
The only solution that comes to my mind, is to copy the hardcoded list to the dynamic persisted array on the very first run and then make accessors for constants etc, to allow that to be changed.

JamesSmartCell:
> @JamesSmartCell as you wrote above about editing of the supplied networks - there are a number of the related hardcoded constants colors/icons/chain_id etc. in the code - it's quite a big task to add editing for that values.
> The only solution that comes to my mind, is to copy the hardcoded list to the dynamic persisted array on the very first run and then make accessors for constants etc, to allow that to be changed.

Ok then, just display the values, don't allow editing for hardcoded networks. Eventually we should refactor this, but not now.

mpaschenko:
@JamesSmartCell added edit of the custom networks

JamesSmartCell:
Works perfectly ! I added the Cronos testnet and transactions updated perfectly and I was able to see my transactions in the explorer.

Just one last thing can you mask the Infura API key in Mainnet, Optimistic etc. The format will always look like ```/[32 hex characters]```, so should be easy to spot.

mpaschenko:
> Works perfectly ! I added the Cronos testnet and transactions updated perfectly and I was able to see my transactions in the explorer.
> 
> Just one last thing can you mask the Infura API key in Mainnet, Optimistic etc. The format will always look like `/[32 hex characters]`, so should be easy to spot.

pushed

JamesSmartCell:
@mpaschenko doesn't work because the Infura key may not be the one that's in BuildConfig.

Can you use a regex for ```/[32 hex characters]```. It would look something like this: ```(/)([0-9a-fA-F]{32})```.

mpaschenko:
updated to regex

JamesSmartCell (approved):

Issue #1967, opened 2021-08-11
Add 'Rate Me' dialog + button


JamesSmartCell:
Need to add the 'Rate Me' button as per just about every other app.

Recommend waiting until 5 app opens, and/or completing a successful action like sending a transaction or receiving some funds.

Therefore add a trigger in WalletFragment, and also in SendActivity.

Check with iOS team under what conditions the dialog appears.

Ensure the dialog looks correct, similar to ones used by apps like Twitter/Quora etc.

mpaschenko:
@JamesSmartCell in a side project I did similar, and logic was:
- 1-3 stars - redirect user to mail form or somekind of the report (which is internal)
- 4-5 start - redirect user to google play, 
to avoid bad reviews...
just FYI.

JamesSmartCell:
Thanks, I was hoping you had previous experience with this! Looks good.

PR #1969, opened 2021-08-12
added rate me dialog (#1967)

mpaschenko:


JamesSmartCell (changes requested):
Can you update it to have a 'material' look:

https://raw.githubusercontent.com/marcoscgdev/RateDialog/master/device-2017-04-14-140649.png

It should have a 'later' button that resets the 5 launches counter (and switches off the 'after transaction' rate).

hboon:
@JamesSmartCell there's also a native `ReviewManager` API which is analogous to iOS' `SKStoreReviewController`. If we can use them, they are usually preferred because users don't leave the app.

JamesSmartCell:
> @JamesSmartCell there's also a native `ReviewManager` API which is analogous to iOS' `SKStoreReviewController`. If we can use them, they are usually preferred because users don't leave the app.

@mpaschenko ^ even better

JamesSmartCell:
@mpaschenko I couldn't get it to work. It correctly calls the inner area ```ReviewManager manager = ReviewManagerFactory.create(context);``` and completes to ```flow.addOnCompleteListener(flowTask -> {``` ...

I'm guessing that the reason it does nothing is because I'm using a build not installed from the PlayStore.
Thing is; we don't know how it would look to an actual user. Does it just dump you straight into the review page without any choice or explanation?

Because there's no way to test it; we should play it safe and revert back to the previous method;

Can you implement the 'Material' dialog as I linked in the picture above, but with five stars across it as seen in other 'rate my app' dialogs. The idea is to make it look attractive to click on; the initial implementation doesn't look good, and this one, we don't know how it looks.

If user clicks on the 'rate' option, or if they click on the stars, then show the stars filled in, then flow to the review page using the ReviewManager you added.

mpaschenko:
@JamesSmartCell  You have to install the build from the app store (or at least from beta) in order to work.

JamesSmartCell:
@mpaschenko yes I thought so. How does it look though? It should be that a small popup appears asking the user to rate the app. If they click on it, then take them there. Allow user to say 'later' or 'no'. Do you know how it appears and what text it uses?

mpaschenko:
@JamesSmartCell it's basically should look like 
Screenshot 2021-08-16 at 8 51 04 AM
but frankly, I was not able to see it, even tried to make my own test app yesterday, and installed it through internal distribution (which should trigger according to manual).
There is some internal logic on if the rate app dialog should be shown, like time quota or so...
https://developer.android.com/guide/playcore/in-app-review
maybe make sense to revert to the initial implementation with an opening the app store page to be sure?

mpaschenko:
> @mpaschenko I couldn't get it to work. It correctly calls the inner area `ReviewManager manager = ReviewManagerFactory.create(context);` and completes to `flow.addOnCompleteListener(flowTask -> {` ...
> 
> I'm guessing that the reason it does nothing is because I'm using a build not installed from the PlayStore.
> Thing is; we don't know how it would look to an actual user. Does it just dump you straight into the review page without any choice or explanation?
> 
> Because there's no way to test it; we should play it safe and revert back to the previous method;
> 
> Can you implement the 'Material' dialog as I linked in the picture above, but with five stars across it as seen in other 'rate my app' dialogs. The idea is to make it look attractive to click on; the initial implementation doesn't look good, and this one, we don't know how it looks.
> 
> If user clicks on the 'rate' option, or if they click on the stars, then show the stars filled in, then flow to the review page using the ReviewManager you added.

@JamesSmartCell would you like me to implement something like that?
image

hboon:
Probably my last bit of contribution to this: there’s a way to test the flow, but it doesn’t display the UI: https://developer.android.com/guide/playcore/in-app-review/test#java

Up to you folks to see if it’s good enough.

mpaschenko:
> Probably my last bit of contribution to this: there’s a way to test the flow, but it doesn’t display the UI: https://developer.android.com/guide/playcore/in-app-review/test#java
> 
> Up to you folks to see if it’s good enough.

I tried as it was described through the internal test track, but was unable to see any UI :(, it could be that with the published version everything would be ok.

hboon:
(If you are referring to the fake version, yes, it doesn't display the UI, it only provides the return value as if the user went through it)

JamesSmartCell:
> > @mpaschenko I couldn't get it to work. It correctly calls the inner area `ReviewManager manager = ReviewManagerFactory.create(context);` and completes to `flow.addOnCompleteListener(flowTask -> {` ...
> > I'm guessing that the reason it does nothing is because I'm using a build not installed from the PlayStore.
> > Thing is; we don't know how it would look to an actual user. Does it just dump you straight into the review page without any choice or explanation?
> > Because there's no way to test it; we should play it safe and revert back to the previous method;
> > Can you implement the 'Material' dialog as I linked in the picture above, but with five stars across it as seen in other 'rate my app' dialogs. The idea is to make it look attractive to click on; the initial implementation doesn't look good, and this one, we don't know how it looks.
> > If user clicks on the 'rate' option, or if they click on the stars, then show the stars filled in, then flow to the review page using the ReviewManager you added.
> 
> @JamesSmartCell would you like me to implement something like that?
> image

Yes I think for now that would be the best solution - combine it with the Playstore rating API you already added - so once the user clicks your 'rate' button in the dialog, they then go to the PlayStore rate dialog you linked to here:

https://user-images.githubusercontent.com/2472117/129517241-d07a620e-fe46-4b6c-861d-4f662046f4f8.png

This is the behaviour I have seen. Listen for the user clicking on the five stars as well as 'rate/confirm' to take them to the dialog. I usually click on the stars in that dialog for some reason ...

mpaschenko:
@JamesSmartCell updated as requested

JamesSmartCell:
Now that I've experienced other apps using the first iteration of this with the bottom sheet, it's actually pretty good. D'oh! We'll go with this one that you done though.

JamesSmartCell (approved):

colourfreak:
I wonder where is this rating going? Is it correlated with Google Play?

Issue #1974, opened 2021-08-16
Update Dialog


JamesSmartCell:
Implement a 'A new version of AlphaWallet has been released. Please upgrade in order to ensure issues are fixed' Dialog.

I think the best way to do this is to check github releases which we maintain alongside each PlayStore release:

```https://api.github.com/repos/AlphaWallet/alpha-wallet-android/releases/latest```

You can pull the latest version ID here: 

```
"node_id": "MDc6UmVsZWFzZTQ3MjIwNTA5",
  "tag_name": "v3.36.1",    <-----------------------
  "target_commitish": "master",
  "name": "Provider connection and stability upgrade",
  "draft": false,
  "prerelease": false,
  "created_at": "2021-08-03T12:03:24Z",
  "published_at": "2021-08-03T12:13:39Z",
```

Compare this with the version on ```BuildConfig.VERSION_NAME```

These are kept in step.

JamesSmartCell:
If you can get the version directly from the PlayStore this is better, but I'm not sure if this is API is available.

mpaschenko:
@JamesSmartCell there is In-App Updates api
https://developer.android.com/guide/playcore/in-app-updates
which can even initiate the update...

JamesSmartCell:
> @JamesSmartCell there is In-App Updates api
> https://developer.android.com/guide/playcore/in-app-updates
> which can even initiate the update...

^ can you implement it?

mpaschenko:
Yes, I had a chance to work with it

PR #1978, opened 2021-08-20
added update dialog (#1974)

mpaschenko:


JamesSmartCell:
@mpaschenko Do you know how this looks, what is the flow when it's in use?

Before we add this I'd like to see how it would look for a user. The way we'd expect this to work is to prompt the user that an update is available, and allow them to dismiss it (not showing them again). If they click that they want to update, then take them to play store.

What would be really bad is if we just jump to the play store when an update is available, or to keep nagging the user to update.

mpaschenko:
viber_image_2021-08-23_09-35-36-281
if an update is available - you would see this kind of dialog (like I made in the test app). You could dismiss it.

mpaschenko:
I can add some delay for the next prompt if a user is refused to update, which number of the days do you prefer?

JamesSmartCell (approved):

Issue #1998, opened 2021-09-06
Icons uploaded to iconssets are not up-to-date in the app


colourfreak:
Maybe caching?

It's seems that there is a conflict between an external library and the one we have in iconssets. 

https://github.com/AlphaWallet/iconassets

colourfreak:
ENJ could be a good example here.

mpaschenko:
@colourfreak  trust wallet asset repo has ENJ token, so first it loads from there:
https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0xF629cBd94d3791C9250152BD8dfBDF380E2a3B9c/logo.png

JamesSmartCell:
@colourfreak we check the trust repo first (because the initial implementation was only a few testnet tokens, since then we've added a huge number of assets).

Do you prefer now to check our icon repo first? If so, @mpaschenko can you re-order the check in class TokenIcon

colourfreak:
@hboon 

Goal: Load up icons provided by us first, if not exist, load up an external library. I manually created and uploaded 140 HQ icons.

colourfreak:
> Do you prefer now to check our icon repo first? If so, @mpaschenko can you re-order the check in class TokenIcon

Yep. The EJN is a good example of how ugly the external graphic can be. 

Since we are all here, let's discuss how to optimize these images. I usually upload a 512x512 px png graphics. That sounds a bit heavy though. Maybe we can think of webp or avif? We can think of lazy loading as well. So let's get image placeholders first (just a gray circle) + name + prices, then load up asynchronously those icons while you are scrolling. 

Thoughts? @hboon @JamesSmartCell @mpaschenko

hboon:
>@hboon  ...

@colourfreak yes. Should be our repo first then fallback to Trust’s then programmatically generated as fallback.

hboon:
I don’t think size is a major issue though. If we want, we can reduce it so it is the largest possible dimension we use it for, eg on iOS is is 3x of maybe 44px (can’t remember offhand the size).

We shouldn’t use other formats because iOS doesn’t do webp and PNG is very performant.

We also don’t need to worry too much about size actually, since the icons ought to be cached locally for a while.

Not worth over-engineering this unless there’s something really bad happening. But we should make loading order, caching etc correct though.

JamesSmartCell:
> > @hboon  ...
> 
[...]

PR #2002, opened 2021-09-08
implemented load of assets from the alpha wallet repo first #1998

mpaschenko:


JamesSmartCell (changes requested):
Can you refactor this in a slightly different way:

Can you restore the code in the original 'displayTokenIcon' and do it this way:

1. Utils.getTokenImageUrl returns the path to AW repo (code should now be much simpler).
2. Fold CircleCrop into the Glide call (first check always AW)
3. If RQ listener gets an 'onLoadFailed' then try to load the Trust icon so you'll need a new function in class Utils as Trust repo needs the chain name, based on the existing getTokenImageUrl. To load the icon, pull the address from the 'model', and match it against the current 'Token token'. If same, proceed, if different then just bail out without an update (RecyclerView reclaimed the icon before it could finish loading).
4. Remove the 'assetSvs.storeImageUrl' from the requestListenerAW onResourceReady, because we'll always be trying AW first so no need for this shortcut. Don't add it into the Trust repo as we want to prioritise our repo first and if the asset is added it should replace the Trust asset.

Note that the assetDefinition.fetchIconForToken in the bindData method is for ERC721 and ERC1155 tokens where we've read the data from the token. Any AW repo icon should override this, so only try this URL after trying AW.

It'll need a little re-code but if it follows the above rules it should be fine.

mpaschenko:
@JamesSmartCell I made changes as you requested, but icons start blinking from time to time **icon <> text** icon, can't get the clue why, will look more tomorrow.

JamesSmartCell:
Thanks, it only needed a small update, can you continue with Samoa - TokenIcon can be a big time sink.

JamesSmartCell (approved):

PR #2037, opened 2021-09-21
Screen size fetch for Android 10 and below vs Android 11.

### LINKED ISSUE #2031: Crash when tap Settings > Show My Wallet Address
Direct APK install. No previous installs.

1 watch wallet only.

v3.4.1 (170).
Huawei nova 7i
Android 10

JamesSmartCell:
Fixes #2031 

For for Android 10 and below API call introduced when deprecated code was refactored.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/BackupKeyActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/BackupKeyActivity.java
@@ -1,22 +1,14 @@
 package com.alphawallet.app.ui;
 
 import android.annotation.SuppressLint;
-
-import androidx.activity.result.ActivityResultLauncher;
-import androidx.activity.result.contract.ActivityResultContracts;
-import androidx.lifecycle.ViewModelProvider;
-
 import android.content.Context;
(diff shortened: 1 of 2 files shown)


hboon:
@JamesSmartCell does it make sense to have a script/Travis to run it on the oldest version of Android supported (in this case <= 10) too? Then either run it with Travis every time or manually occasionally?

JamesSmartCell:
> @JamesSmartCell does it make sense to have a script/Travis to run it on the oldest version of Android supported (in this case <= 10) too? Then either run it with Travis every time or manually occasionally?

Yes that would be very good. I'm thinking at the very least we can build a series of macros that test all the wallet functions, perform actions etc. I also recall we used to do random monkey test suites that click all over the place.

hboon:
Yes, possibly. There's varying granularity, but a lower hanging fruit might be just to display certain screens (which already covers this case as well as loading/refreshing tokens, etc) and then moving on to calling certain functions/flows and snapshot tests.

Anyway, not to pollute your PR, but something for you to think about.

PR #2087, opened 2021-10-05
Fix signed transactions count for WalletConnect session doesn't updat…

### LINKED ISSUE #2083: The number of signed transactions for WalletConnect session doesn't update right after signing
1. Connect https://example.walletconnect.org
2. Click eth_sign (or the 712 version)
3. Count in screen still says 0 until we go back and open the screen again.

seabornlee:
Closes #2083 @JamesSmartCell @hboon


JamesSmartCell (approved):
Elegant solution, nice.

I edited your comment adding 'Closes' to prefix the issue number, then it auto-closes the issue when merged.

PR #2165, opened 2021-10-20
implemented exchange with 1inch #2124

### LINKED ISSUE #2124: Oneinch should be available for the native crypto token on Ethereum mainnet and Arbitrum One
Refer to https://github.com/AlphaWallet/alpha-wallet-android/issues/2061#issuecomment-929776367

mpaschenko:
Closes #2124


JamesSmartCell:
Hi @mpaschenko this one was evil! The best I can determine the 1Inch website has a background watchdog thread that we were fighting with.

The heuristic that finally worked 100% of the time was this:

```switchNetworkAndLoadUrl```

- Store the chainId and Url in members
- update the Web3 injection data
- reload the page, and listen for page reload complete

```public void onWebpageLoadComplete()```

- after reload is complete, check if we set the URL, if so, start loading that page
- Set a timer to blank the chain change for 3 seconds (the watchdog won't always be triggered)

```OnWalletAddEthereumChainObject```

- if chain change dialog is triggered by the watchdog with 3 seconds of page load, ignore it.

...
3 seconds passes
...
timer triggered to clean up forceChainChange


It's a bit nasty, but considering how the multi-chain seems to work it's the best I could come up with.

PR #2209, opened 2021-11-03
2196 fold the ethereum custom chain structure

### LINKED ISSUE #2196: Fold the Ethereum Custom Chain structure into networkMap
Currently there's a lot of extra handling for the custom chains, which leads to some duplicated code and a few 'gotchas', both now and in the future.

Can you add a 'isCustom' boolean flag to NetworkInfo, and then, on initialisation add the custom chains in - no danger of adding them multiple times because it's a SparseArray.

Then, simplify the rest of the handling in EthereumNetworkBase accordingly.

mpaschenko:
Closes #2196

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -186,7 +186,7 @@ dependencies {
     implementation "com.google.dagger:dagger-android:2.22"
 
     //Use Leak Canary for debug builds only
-    debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.7'
+    //debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.7'
 
     // Image Loader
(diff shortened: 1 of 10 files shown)


JamesSmartCell:
@mpaschenko this looks a lot better, but can we take it one step further and completely remove the NetworkExt? Suggest you add the nets with real value (ie not testnet) into the ```hasValue``` list (checking for duplicate entry). That way it streamlines the testnet checking too, and adds order for the new custom networks.

mpaschenko:
Updated

JamesSmartCell:
Looks great now, will have a final test tomorrow and merge. This will help avoid future situations where we add a new function involving known chains and forgot to add/check the custom user chains.

JamesSmartCell (approved):

PR #2240, opened 2021-11-21
Rename element to avoid duplication names in 3 layout files, it may c…

seabornlee:
…ause

ClassCastException

Different types have the same name.
image

image

I think have a type prefix is good name convention, like the `text_` below .
image


JamesSmartCell:
@seabornlee thanks :) I had already spotted this one and fixed independently.

seabornlee:
> @seabornlee thanks :) I had already spotted this one and fixed independently.

Great!

PR #2259, opened 2021-12-04
E2e test

seabornlee:
Can you review and merge the test first, so that everybody can run it locally if need.
I will make it automatically run on Travis in next PR.


JamesSmartCell:
Hi @seabornlee, will do, once the tests pass.

It's going to be a great improvement once we have these auto-tests running. I can see these getting extended quite a bit too.

seabornlee:
The app still can not get started in Travis emulator, I just undo my .travis.yml change.
As we can only have one travis job running at the same time, I will try to debug the build on my own fork so it won't stuck the other builds.

JamesSmartCell:
Hi @seabornlee the currently working setup; does this include the emulator performing tasks? I couldn't work out which part of the build feed is where this is happening.

Once we have this it's going to be really good - I would like to run through the tests not only on an API30/31 emulator but also an API24 emu.

seabornlee:
> 
I tried running on API 24 emulator, it works on my MacBook arm emulator, but it crashed on Travis CI, I need to get into debug mode to figure out the reason.
The current travis config not run e2e test, because it will fail.
I suggest we can separate the test and CI configuration, so every body can run or enhance the e2e test suite locally if they want.
I will continue trying to make it working on CI server.

JamesSmartCell:
Hi @seabornlee are the new tests ready to be merged into the main repo? It all looks good to me.

seabornlee:
Not ready, very close. Now I am running it on API level 24, 30, and 31, randomly fail on one of them. Still trying to make test stable.

seabornlee:
Close this and will create a new one later.

Issue #2233, opened 2021-11-18
Redesign ERC721 to be align with recent ERC1155 update we made [ERC721]


colourfreak:
The good news is, that we already made the layout for ERC1155, which now can be applied to the ERC721 redesign. Note, that unlike ERC1155, in the Assets tab you will not find the menu icon in the top right. Since there no way for multiple selection, I skipped it. 

**Zeplin:**
https://zpl.io/bzzgow7

![ERC721-android](https://user-images.githubusercontent.com/51817359/142425180-b31b52ef-30eb-4708-8866-19f2a93911ab.jpg)

Another Example is with FIFA tickets:
![ERC721-fifa-android](https://user-images.githubusercontent.com/51817359/142425683-c904c093-94b1-48f3-b5fd-225c4bcaced7.jpg)

https://zpl.io/agEY0nR

https://user-images.githubusercontent.com/51817359/142426556-8be43902-ea56-4741-b0c9-1f77eadca9db.mov

JamesSmartCell:
Nice, this would be a great task for @justindg .

Just one clarification point, when you say 'Since there no way for multiple selection, I skipped it.' you're referring to ERC721 explicitly?

![screen1](https://user-images.githubusercontent.com/12689544/142782732-6ddba4b1-80a6-4f42-80ec-234252ce77d1.jpg)
![screen2](https://user-images.githubusercontent.com/12689544/142782739-9eec4a82-3c28-4072-a01b-8c4c4238c5f4.jpg)
![screen3](https://user-images.githubusercontent.com/12689544/142782765-772789ee-ec13-4737-8caf-fd3e6701de29.jpg)

Because we support this for ERC1155 (as per your original spec).

hboon:
> Just one clarification point, when you say 'Since there no way for multiple selection, I skipped it.' you're referring to ERC721 explicitly?
> Because we support this for ERC1155 (as per your original spec).

Yes, this issue is only for ERC721; to make the UI for ERC721 look similar to ERC1155's.

colourfreak:
What's the status of this? Who is working on this? Who manages this task? Shall we re-assign? @hboon

justindg:
> What's the status of this? Who is working on this? Who manages this task? Shall we re-assign? @hboon

I have opened a pull request last month. Most of the work is done, and there are some minor issues that needed to be addressed first before we can merge. 

https://github.com/AlphaWallet/alpha-wallet-android/pull/2271

Issue #2232, opened 2021-11-18
Add Search box under the segmented control inside Assets tab [ERC1155]


colourfreak:
It should work natively. When you tap on it, the search box appears at the very top.
https://zpl.io/bW3ogpk
https://zpl.io/b6Wzjkm
https://zpl.io/293QEpe

![ERC-1155-Project-Assets-Grid](https://user-images.githubusercontent.com/51817359/142417098-a93977f3-7bd4-4084-8609-1f936bc602c6.jpg)

![ERC-1155-Project-Assets-List](https://user-images.githubusercontent.com/51817359/142417100-4f065cb1-5e18-4351-b990-69f929b7c949.jpg)

![ERC-1155-Project-Assets-Search-Results](https://user-images.githubusercontent.com/51817359/142417678-7556ce4b-aecd-47be-86d2-02c6b0d6eb35.jpg)

JamesSmartCell:
@justindg can you add this (After Samoa goes in). I assume the search bar is within a view holder on the TokensAdapter, so it should drop into the 1155/721 view.

This should go in straight after the work to merge the 721 code into 1155.

justindg:
Solved in #2271

PR #2271, opened 2021-12-09
ERC721 Redesign

justindg:
PR contains:
- Redesign ERC721 display - #2233 
- Implement grid view for NFT assets
- Bug fix: NFT images not loading properly
- Minor enhancements to NFTImageView scaling
- Search functionality - #2232 
- Added temporary working info for NFTInfoFragment; AssetContract doesn't seem to work for now
- Merge ERC721 & ERC1155 classes to NFTxxx classes where applicable - #2277

Further details:
- All ERC1155xxx classes renamed NFTxxx EXCEPT for ERC1155AssetList / AssetSelect (as they are specifically for ERC1155 tokens)
- ERC721 tokens like CryptoKitties are now displayed similarly to ERC1155 tokens, which is through the new NFTAssetsAdapter (previously NonFungibleTokenAdapter). OpenseaHolder class is now removed as well.
- All other types ERC721 tokens such as tickets & ENS (which use the AssetInstanceScriptHolder) are untouched, and are still displayed through NonFungibleTokenAdapter. In NFTAssetsFragment, I have written a function to check if a TokenScript override is needed, so that the RecyclerView can use either NonFungibleTokenAdapter or the new NFTAssetsAdapter.

Thoughts:
- In the future, NonFungibleTokenAdapter should probably be removed entirely. I think there's too much going on in this class. This is probably possible once the TokenScript stuff is merged to the new NFTAssetsAdapter, but currently I don't have much knowledge on this area yet.


JamesSmartCell:
Can we use abstraction to cut down on the number of new classes? I would imagine most of the assetview and view models would be the same as those for ERC1155.

JamesSmartCell:
Further suggestion:

I think the essence of what we wanted was to take the existing ERC1155Activity/ERC1155AssetDetailActivity ..., etc

Rename it NFTActivity, NFTAssetDetailActivity, and abstract the display of ERC721 and ERC1155 with minimal changes to these classes.

Would it be possible to design it this way?

For example, pretty much all the code in the ERC1155 classes will be unchanged, but we would exclude the ability to do multi-select (these could be abstracted into the ERC721Token and ERC1155Token classes).

This way we have a much more compact update.

justindg:
@JamesSmartCell there are some issues that I may need your help on:

1) In the ERC721 token subtitle, the description indicates `Fungible Token`, I think there may be issues with the classifier
image

2) The ENS display is a little wonky with the Grid View, unsure how to approach tokens like this
image

3) The ENS tokens actually place nicely if I slot them in the new NFT Assets Adapter, BUT for some reason one of the tokens doesn't show up properly (the assetMap is null somewhere in the contract). I used Tomek's wallet via watch-wallet for this.
image

JamesSmartCell:
@justindg seems to be going well - once you're ready to continue we'll need to go through and ensure that all the TokenScript functionality from AssetDisplayActivity is still working. We'll probably need a further abstraction to make sure it's working - there are a lot of nuances due to displaying webviews in the RecycleView.

JamesSmartCell:
> @JamesSmartCell there are some issues that I may need your help on:
> 
> 1. In the ERC721 token subtitle, the description indicates `Fungible Token`, I think there may be issues with the classifier
>    image
> 2. The ENS display is a little wonky with the Grid View, unsure how to approach tokens like this
>    image
> 3. The ENS tokens actually place nicely if I slot them in the new NFT Assets Adapter, BUT for some reason one of the tokens doesn't show up properly (the assetMap is null somewhere in the contract). I used Tomek's wallet via watch-wallet for this.
>    image

There's no need to show fungible/non fungible for ERC721, since these can only be Non Fungible.

JamesSmartCell:
Hi @justindg you may already know this; the ERC1155 token instances can be Fungible, Non-Fungible or Semi-Fungible. ERC721 is always Non-Fungible and is always a count of 1. The '1 Asset' for ERC721 is kind of redundant but if it's in the spec then we should leave it.

justindg:
> Hi @justindg you may already know this; the ERC1155 token instances can be Fungible, Non-Fungible or Semi-Fungible. ERC721 is always Non-Fungible and is always a count of 1. The '1 Asset' for ERC721 is kind of redundant but if it's in the spec then we should leave it.

Got it. I'd try to tweak the code instead. The problem was it displayed '0 Assets' for the 721 tokens, also the method `getAssetCategory()` seemed to always return 'Fungible Token'

JamesSmartCell:
@justindg let me know when this one is complete, can you mark it with 'draft' until it's ready?

justindg:
@JamesSmartCell this should be complete already, at least it works for Watch-only wallets.

JamesSmartCell:
Awesome, I'll give it a more detailed look

JamesSmartCell (changes requested):
I have a few issues I found on the first iteration:

1. When you click on the NFT (either ERC721 or ERC1155) in the wallet view you go to NFTActivity. Both are handled by separate routers - can you remove AssetDetailRouter and move the method into TokenDetailRouter.
2. In NFTActivity, For ERC721 I'm seeing each asset (in grid view) has incorrect Assets being displayed: 
Screenshot_20220122-121537_AlphaWallet
You only need show the Title here for ERC721, no need to show 6 Assets (should be 1 asset), and no need to show NFT, since we know this is individual NFT tokens. At least, only show 1 Asset | NFT, but that is redundant.
3. the 'Info' tab should only be displayed if we have overall Token info from Opensea, or read from the Contract Metadata. Currently we don't do this, so Info should be hidden for the first release.
4. TokenScript buttons don't seem to work correctly (but we can go into these later).
5. The loading views need to be protected at the glide image load; in the case of when images were taking a long time to load I clicked back, and if they finally were loaded then the app crashes as the view has been destroyed. Mostly crashing at 'fetchAsset' Line 118 NFTAssetsAdapter.
6. Selection view for ERC1155 doesn't quite look right - I'll revert and show a before/after.
7. Asset count for ERC1155 is incorrect.

Here's a watch wallet:

0xA20efc4B9537d27acfD052003e311f762620642D

Watch Rinkeby on this account, check out the 'Chicken Giblets' NFTs which are ERC1155.

Here's before:

Screenshot_20220122-123650_AlphaWallet

And after:

Screenshot_20220122-123416_AlphaWallet

Somehow the 'Fungible Tokens' has become 'NFT' which is incorrect. This may be my bug not sure. You can see the counts for each token is now incorrect.


When I go to select, before:
Screenshot_20220122-123938_AlphaWallet

After:

Screenshot_20220122-123425_AlphaWallet

The thumbnails are distorted (aspect ratio) and we've lost the first few letters of the detail (Maybe check if the spec has changed, but it used to have that detail).


A few times, if I went to one NFT and into NFTActivity, then came out, and went to a different NFT (NFTActivity) I would see some image weirdness eg the image was taking the whole template rather than the top of the view as normal. Also, the load indicators don't seem to be working, or work inconsistently.

Have a play with the watch wallet I copied above, in Rinkeby, and you'll see what I mean.

I'll send you the TokenScript corresponding to the message tokens too; then you can see the buttons not working. This is most likely because for Action-only script they weren't working properly when you started this mod! I fixed them in December.

justindg:
@JamesSmartCell can you have a look at this branch? I pushed several changes and so far it seems stable.

Regarding "image weirdness", what I did was for the images to fill the containing layout - centered and preserving aspect ratio. I had to find a compromise as there is a variety of image sizes in different orientations as well, which I have to fit in a square-ish grid.

Let me know what you think! I will continue to test as well.

JamesSmartCell:
The images are still distorted compared to the current implementation in master - I don't think we should do any scaling to any images, like stretching or squashing - if they have to appear not as square it's ok. I found constraining vertically gives the best result.

I got an image load crash but this was my fault in old code - I'll fix those correctly! (rather than simply checking the activity isn't destroyed or being destroyed).

The NFT listview for ERC1155 isn't working correctly as it was previously, please see this zeplin:

https://app.zeplin.io/project/5d088205bff2d15de6a4397b/screen/619628f0620aedaa439b3409

You can check out the old ERC1155 layout which we both worked on (which may need tidying up). It should show more information both in the gridview and in the layout view (as previous).

ERC721 doesn't need the extra info in grid view.

You can watch Rinkeby 0xA20efc4B9537d27acfD052003e311f762620642D for an example with lots of ERC1155 (Chicken Giblets Token) which should show Semi-Fungibles.

Watch Tomek's account to see ERC1155 Non Fungibles and Semi-Fungibles (you may need to refer to the current master branch which renders these correctly).

Also - can you use a placeholder progress spinner when entering the NFT detail (ie, I click on an NFT in the grid or list view), if the image hasn't loaded yet; the text springs down once the image loads.

Other than these it looks good and the code is definitely cleaner than when we hacked these in previously :)

Appreciate you using the regular brace pattern too! I should check that into the settings.

That is this:

```
        if (item != null && item.requiresReplacement())
        {
            fetchAsset(holder, pair);
            if (!holder.icon.hasContent())
            {
                holder.loadingSpinner.setVisibility(View.VISIBLE);
            }
        }
```

rather than Egyptian:

```
        if (item != null && item.requiresReplacement()) {
            fetchAsset(holder, pair);
            if (!holder.icon.hasContent()) {
                holder.loadingSpinner.setVisibility(View.VISIBLE);
            }
        }
```

Which I find slightly less pleasant to look at.

justindg:
> The images are still distorted compared to the current implementation in master - I don't think we should do any scaling to any images, like stretching or squashing - if they have to appear not as square it's ok. I found constraining vertically gives the best result.

I haven't done any scaling or stretching for any of the images. Essentially, what I did was to "cut" a square area from the center of the image like below (red square being the displayed area):

image

I found that this is the best compromise for a few reasons:

- The layout will be consistently displayed as a square, so there are less surprises no matter the image aspect ratio
- Extremely long images AND horizontal images will not display pleasantly. The Chicken Gibblet NFTs had a decent proportion so they appeared okay
- Since we are reusing NFTImageView for both list and grid views: if the image is constrained vertically, the grid display will not be 'filled out' for vertical images - leaving a blank space at the sides

Example of a very long vertical NFT:
photo1644428088 (3)


> The NFT listview for ERC1155 isn't working correctly as it was previously, please see this zeplin:
> 
> https://app.zeplin.io/project/5d088205bff2d15de6a4397b/screen/619628f0620aedaa439b3409
> 
> You can check out the old ERC1155 layout which we both worked on (which may need tidying up). It should show more information both in the gridview and in the layout view (as previous).

I intentionally hid the subtitle/description if the `getAssetCategory()` returns `NFT`. Below is an example from Tomek's wallet with mixed asset types:
photo1644428088 (4)

> ERC721 doesn't need the extra info in grid view.

Seems fine on my end? Same reason as above - hid subtitles if `getAssetCategory()` returns `NFT`. Example below:

photo1644428088 (1)

> Also - can you use a placeholder progress spinner when entering the NFT detail (ie, I click on an NFT in the grid or list view), if the image hasn't loaded yet; the text springs down once the image loads.

I could add this to the detail layout so the user would expect something to load in that area, but if the image is too long, the 'springing' effect will still happen. To alleviate this is to set a minimum height of 350dp to the area + add the progress bar. Is this an acceptable compromise?

--

@JamesSmartCell Let me know your thoughts on this. This feature is very tricky with lots of edge cases, so please bear with me. Probably best to loop in @colourfreak as well if he has any input.

JamesSmartCell:
@justindg thanks for the update - since then I took a closer look and saw that you're not re-scaling, you're cropping which like you say is the best compromise, so please keep that.

So in summary I think the changes we need are:

- display ERC1155 correctly (with asset count, NFT or Semi-fungible)
- have a loading spinner for the detail page to indicate something will happen, and have a placeholder space.
- (@JamesSmartCell fix long image loading crash)

I'm divided over the full width picture rendering for the detail page - I like that I can see more detail (how you have it now, the pictures look great), but it takes up too much room and if they are long-loading the page jumps a lot on load. Maybe have a fixed height for the detail page but if user clicks on the image then show it full screen? I'll leave that to @colourfreak

JamesSmartCell:
@justindg I just checked Tomek's erc1155 and those render correctly so the problem looks like it's on my end.

colourfreak:
@justindg Very smart idea. We are dealing with Collection's cover images on iOS as well. 

So if I go your way right, you take the middle square out of the very long cover. 
Let's use this as an example:
https://opensea.io/collection/zoofrenznft

The cover looks like this
Screenshot 2022-02-10 at 12 30 54

[Link](https://lh3.googleusercontent.com/_YXZRyJyQHpd0mxE35F4np2SttVbV44__ayQzXef9Zn5ZLNqROUF7Ea2YqYNoJOoKjUXp1OIM9s1_AWR4JoA0nGt51LEDvYCujtP=h600)

So does it mean that using API, you take the middle square and place it as a collection's cover?
Screenshot 2022-02-10 at 12 30 19

hboon:
@colourfreak for images from OpenSea, refer to: https://github.com/AlphaWallet/alpha-wallet-ios/issues/3894. Search for "smart"

JamesSmartCell:
@tomek this PR is ready for merge, can you create a separate issue to cope with non-standard shaped images/

JamesSmartCell (approved):

Issue #2218, opened 2021-11-15
Migrate Paging to ViewPager2


JamesSmartCell:
Migrate main page adapter (Wallet, Activities, DappBrowser, Settings) in HomeActivity to use ViewPager2, as ViewPager has been deprecated.

See the examples here:

https://developer.android.com/training/animation/screen-slide-2
https://www.geeksforgeeks.org/viewpager2-in-android-with-example/

asif-finimble:
Task Update Jan 17, Monday:   
Migrated to ViewPager2 in HomeActivity.     
         
Current implementation uses custom ```ViewPager``` called ```ScrollControlViewPager``` which does 2 additional things:  
- 1] ```lockPages(true)```. I think this is to disable user Touch input or swiping across pages.      
- 2] uses PagerCallback to notify ```loadingFinished()``` on adding last Page of homescreen. Page No = ```WalletPage().values.length```  .      
      
I replicated the above 2 as follows:   
- 1] using ```viewPager2.setUserInputEnabled(false); ```.  Disables swiping.       
- 2] Settings fragment is the last page that is created in Adapter.  Call ```loadingFinished()``` after that.      


Note: According to documentation, Its mandatory to create fragments in adapter instead of returning already existing instance. So instead of ```getItem()``` updated code has ```createFragment()```.      

  
```ERC1155Activity``` WIP

asif-finimble:
Task Update Jan 18, Tuesday:   
Migrated to Viewpager2 in ```ERC20DetailActivity``` and ```ERC1155Activity```.   
- Tested ```ERC20DetailActivity```.   
- Don't know how to test ```ERC1155Activity``` but should work fine.    
- ```ImportWalletActivity``` is crashing on migrating to ViewPager2 due to the fragments not able to access parent activity views.

asif-finimble:
Task Update Jan 19, Wednesday:   
 - Fixed by using ```getView().findViewById()``` instead of ```getActivity().findViewById()```  in fragments.   
 - Done. Pushed the code.   
   
  
@JamesSmartCell, could you please go through the comments and let me know if the approach I used needs modifications? 
Also I apologize but I also migrated other usages. It got out of my mind that we just need to migrate main page. Should I revert others except main page?.

PR #2383, opened 2022-01-31
Migrated ViewPager to ViewPager2

asif-finimble:
Closes: #2218


JamesSmartCell (changes requested):
Please avoid using numbers in your variable/object names; we do sometimes use numbers at the end of the variable/object names but this is for when we are replacing a legacy class; once the legacy class has been removed the number would be removed.

Eg: GasService2 was in place for a while while GasService was being deprecated.

asif-finimble:
Merged master and resolved conflicts.

JamesSmartCell:
@asif-finimble Expected merge conflict from your fragment fix

asif-finimble:
@JamesSmartCell the conflicts are resolved.

JamesSmartCell (approved):

PR #2417, opened 2022-02-19
Replaced WalletConnect request popup with ActionSheet.

### LINKED ISSUE #2228: Make WalletConnect signature request be clearer about where it's coming from #3289
Refer https://github.com/AlphaWallet/alpha-wallet-ios/issues/3289#issuecomment-953465632

asif-finimble:
Closes #2228.


JamesSmartCell:
Ok. I probably should have explicitly said how to implement this. I did do a writeup on how I would like this done in the issue, but I'll break it down more finely here, as possibly it wasn't clear.

- Please, no Kotlin classes unless discussed. Mixture of Kotlin and Java looks messy. We only use Kotlin for some WC because that's part of the sample implementation and @justindg  and I conferred about adding those. 
- The coding style for the ActionSheetDialog is using widgets, so the chain request should be self contained within a widget. Create something like 'WalletConnectRequestWidget'.
- You can roll your DialogInfoItem into the widget mentioned above.
- Use the existing ActionSheetCallback interface.

This way, most of the new code can be rolled into the WalletConnectRequestWidget

The final solution should look something like this:


```
confirmationDialog = new ActionSheetDialog(peer.getUrl(), chainIdOverride, icon, this(for ActionSheetCallback), etc );
confirmationDialog.show();
```

You would then pass these params to the WalletConnectRequestWidget widget from ActionSheetDialog (new constructor)
The extra button function (ie Connect/Cancel) can be added using member class functionBar.

see: https://app.zeplin.io/project/5d088205bff2d15de6a4397b/screen/6012c08d598d3aaf4482cdb7 (but ignore the colouring for now, that is being done separately).

You may need an extra path in the ActionSheetCallback interface

At a future date, in a separate PR we will probably separate out the message and request types from ActionSheetDialog and start a new RequestSheetDialog class.

The code you wrote does look good and solid, but having a separate ActionSheetDialog is a little confusing in terms of design.

JamesSmartCell (changes requested):
Please see design notes in the issue and expanded in the thread.

asif-finimble:
Hi @JamesSmartCell thanks for the detailed explanation.   
- I did use the function bar but was not able to get the rounded buttons as shown in the related IOS issue. https://github.com/AlphaWallet/alpha-wallet-ios/issues/3289#issuecomment-952859643.     
- I will update the code accordingly.

asif-finimble:
@JamesSmartCell I have updated the code to use ActionSheet. Let me know if I need to make any change.

JamesSmartCell:
> @JamesSmartCell I have updated the code to use ActionSheet. Let me know if I need to make any change.

Thanks for the refactor, yes this definitely looks like what I was expecting! It fits with the existing flow. Later we'll split ActionSheetDialog as it's starting to get a bit too full.

asif-finimble:
Yes splitting the ActionSheetDialog will be good. I think this can be merged now. Let me know once you test it and if any changes are required.

JamesSmartCell (approved):

PR #2418, opened 2022-02-23
Parcelised WalletConnectRequest.

### LINKED ISSUE #2091: Parcelise the WalletConnect data objects and send in the service Intent
See WalletConnectService and WalletConnectViewModel comms

The flow is like this:

1. Message from WalletConnect (via https)
2. WalletConnectService picks up the message, broadcasts an intent that message was received ```Intent(WALLET_CONNECT_REQUEST);```
3. ```WALLET_CONNECT_REQUEST``` contains the sessionId key, is picked up by WalletConnectActivity ```walletConnectActionReceiver```
4. WalletConnectViewModel connects to WalletConnectService and fetches the WCRequest object.

By Parcelising WCRequest (or the required subset of it) we can transmit the data object directly, and steps 3 and 4 become:

3. ```WALLET_CONNECT_REQUEST``` contains the WCRequest, is picked up by WalletConnectActivity ```walletConnectActionReceiver``` which pulls the WCRequest out of the intent and actions the message directly.

asif-finimble:
closes #2091


JamesSmartCell:
@asif-finimble this looks great, will test.

JamesSmartCell:
Thanks to your update I was able to tidy up some old code. The change I made was to also send the request intent when an action is sent, and to remove the action from the queue if we receive it directly. Can probably refactor ```getPendingRequest``` out now, because the message pump in the service should ensure WC gets the message.

Can you:

1. Deduplicate the method calls in ```BroadcastReceiver walletConnectActionReceiver```
2. Investigate the TODO I added line 285 286 of ```WalletConnectService```. There should be a way to determine if the app is running and in foreground from the service itself (hopefully without statics), if so then no need to add the request to the stack. The reason we add the request to the stack is if the user is running the dapp from a browser within the phone.

You can test this by going to: jamessmartcell.github.io in eg brave browser/chrome etc on your mobile, connect to wallet. This will result in the broadcast not being delivered, so we add it to the queue, then when AlphaWallet is paged back in the broadcast finally gets picked up and we pull the request from the intent. If we know if the broadcast is going to be seen or not before we broadcast we can simplify the code a bit.

Does it make sense?

JamesSmartCell (changes requested):
Please see above - it's not any change to your code but I saw an opportunity to clean up some old code after you made this.

asif-finimble:
Sure. I'll take look and update the code.

asif-finimble:
@JamesSmartCell I tried applying the suggested changes. I don't exactly understand the TODO as the App will be in running state at that point.   

I have sent broadcast from HomeActivity for foreground status and receive it in WalletConnectService and do a messagePump/checkMessages once.  But this will happen everytime HomeActivity comes into foreground, and not the App.   

Let me know if I missed something. Can you review and tell me if this is what you want?

JamesSmartCell:
Sure, will check; yep it's fine to do the message pump everytime HomeActivity comes into foreground. The flow this is solving is described above:

> You can test this by going to: jamessmartcell.github.io in eg brave browser/chrome etc on your mobile, connect to wallet.

The wallet is paged out when using a mobile browser. If I use WalletConnect from the mobile browser it will open AlphaWallet. If I page back to the browser and fire a WC command I have to page back to AW so the command can be sent from the service.

This is in hand with the WC refactor but for now (we will use a notification service), pumping the message from when AW is paged back in should solve the issue, unless two consecutive commands were sent! Dapps should not allow that though; they should have a 'waiting' dialog.

asif-finimble:
Yes I have already tested the github page and the wallet connect request is shown and working if we approve/reject.

JamesSmartCell:
> Yes I have already tested the github page and the wallet connect request is shown and working if we approve/reject.

Thanks, can you roll your broadcast receiver into the 'onStartCommand' and communicate to the service in the same way as in ```WalletConnectViewModel``` for consistency; the service might have gone inactive; it's best to communicate eg like this:

```
    public void rejectRequest(Context ctx, String sessionId, long id, String message)
    {
        Intent bIntent = new Intent(ctx, WalletConnectService.class);
        bIntent.setAction(String.valueOf(WalletConnectActions.REJECT.ordinal()));
        bIntent.putExtra("sessionId", sessionId);
        bIntent.putExtra("id", id);
        bIntent.putExtra("message", message);
        ctx.startService(bIntent);
    }
```

except create a new member to WalletConnectActions. This ensures the service is running correctly and avoids creating new broadcast services.

asif-finimble:
Applied the changes.

JamesSmartCell:
No need for the new BroadcastReceiver. Can you communicate with the service in the same style as per eg `rejectRequest` above. So, you'd need to add a new command to the `enum WalletConnectActions`, and then send this via the same way as per rejectRequest above, so `broadcastStateToWCService` would have a call to `startService`.

asif-finimble:
@JamesSmartCell I have applied the changes and cleaned the code a bit.

JamesSmartCell (approved):

PR #2441, opened 2022-03-10
Update DAS Dapp info

Mancy009:
DAS has rebranded as .bit. And the new website is https://did.id


JamesSmartCell (approved):

Mancy009:
Hi, James Brown. Sorry for the late reply! I just saw the email.
The old one can be used as well. But we suggest that you can update to the
new one: indexer.did.id

PR #2452, opened 2022-03-21
Allow edit built-in network

### LINKED ISSUE #2347: Allow Override of RPC and chain details for known chains
Currently the 'chain edit' doesn't allow you to edit known chains; it's come to our attention that some users may want to do this.

The edit screen (Settings -> Select Active Networks -> Hamburger -> View) is read only for these chains. For known chains, we should be able to edit everything except the chain ID.

Also, there should be a 'reset to default' button for these chains.

seabornlee:
Close #2347

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/AddCustomRPCNetworkActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/AddCustomRPCNetworkActivity.java
@@ -4,16 +4,12 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.text.InputType;
 import android.text.TextUtils;
-import android.view.View;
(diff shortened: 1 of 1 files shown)


JamesSmartCell (changes requested):
Would it be straightforward to be able to restore the default settings if you edit them?

seabornlee:
How about adding a `Reset` button to restore the default value of a network?

JamesSmartCell:
> How about adding a `Reset` button to restore the default value of a network?

Yes, that would be perfect. People could accidentally change the values, or want to try a different RPC, but then think it's better with the built-in ones. They'd have no way to do this other than re-installing.

seabornlee:
> They'd have no way to do this other than re-installing.

Got it.

PR #2460, opened 2022-03-22
List apk on fdroid

### LINKED ISSUE #1573: Howcome Alphawallet is not listed on F-droid?
Would you be able to add this on F-droid please? Some users arent comfortable downloading from playstore.

seabornlee:
Close #1573


app/build.gradle
```diff
@@ -223,6 +204,12 @@ dependencies {
 
     //Timber
     implementation 'com.jakewharton.timber:timber:5.0.1'
+
+    //Analytics
+    analyticsImplementation 'com.google.android.play:core:1.10.2'
+    analyticsImplementation 'com.google.firebase:firebase-analytics:20.0.0'
+    analyticsImplementation 'com.mixpanel.android:mixpanel-android:5.8.4'
+    analyticsImplementation 'com.google.firebase:firebase-crashlytics:18.2.5'
 }
```
JamesSmartCell, line 213: I don't think we want to put these here. The idea behind a 'noAnalytics' flavour was specifically to avoid including these libraries when making the build.

Are they excluded somewhere else? Eg when making the build does the Java build engine exclude unused files?

JamesSmartCell (commented):
Not sure about moving the Google libraries out of the 'analytics only' folder and into the global space.

app/build.gradle
seabornlee: analyticsImplementation means gradle assembleAnalytics will include Google libraries, gradle assembleNoAnalytics won't include.
For Chinese app market, we can use noAnalytics build, am I right?
[8:34 AM]
https://riptutorial.com/android/example/402/adding-product-flavor-specific-dependencies
Android Tutorial => Adding product flavor-specific dependencies
Learn Android - Adding product flavor-specific dependencies
[8:34 AM]
This is an example.

JamesSmartCell:
@seabornlee makes sense, I get it now, very elegant solution.

JamesSmartCell (approved):

PR #2470, opened 2022-03-24
Added walletconnect switchEthereumChain and addEthereumChain support. 

### LINKED ISSUE #2404: Support both wallet_switchEthereumChainObject and wallet_addEthereumChain JSON-RPC methods in WalletConnect v1
Context:

Pedro:
> This would depend on the wallet… walletconnect allows any JSON-RPC methods

https://canary.discord.com/channels/492410046307631105/833628540456140821/905050546904399892

Reproduce:

https://github.com/AlphaWallet/alpha-wallet-ios/issues/3887#issuecomment-1034527245

asif-finimble:
Closes #2404

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/WalletConnectActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/WalletConnectActivity.java
@@ -53,6 +53,7 @@ import com.alphawallet.app.walletconnect.entity.WCEthereumTransaction;
 import com.alphawallet.app.walletconnect.entity.WCPeerMeta;
 import com.alphawallet.app.walletconnect.entity.WalletConnectCallback;
 import com.alphawallet.app.web3.entity.Address;
+import com.alphawallet.app.web3.entity.WalletAddEthereumChainObject;
 import com.alphawallet.app.web3.entity.Web3Transaction;
 import com.alphawallet.app.widget.AWalletAlertDialog;
 import com.alphawallet.app.widget.ActionSheetDialog;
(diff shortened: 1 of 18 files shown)


JamesSmartCell (changes requested):
Looks good! Just one thing, Where you have the onSwitchChainRequest this should be an ActionSheet request. See 'showChainChangeDialog' in class DappBrowserFragment.

And, the string is available translated as 
```String message = getString(R.string.request_change_chain, newNetwork.name, String.valueOf(newNetwork.chainId));```

Together with a warning if the change chain request is switching between testnet/mainnet as per the same method above.

With this small change it looks all good, nice work.

**Comment by asif-finimble** [comment]
@JamesSmartCell updated the code to use action sheet same as dappbrowser fragment.

File: app/src/main/java/com/alphawallet/app/ui/WalletConnectActivity.java [hunk 1]
```diff
@@ -1216,11 +1254,138 @@ public void denyWalletConnect()
...
+                    switchChainDialog.show();
+                    switchChainDialog.fullExpand();
+                }
+                catch (Exception e)
+                {
+                    Timber.e(e);
+                }
+            }
+        }
+    }
```
JamesSmartCell, line 1328: Can you refactor this method and split it up a little - maybe perform all the validation in a separate method. that way we can at least get rid of one layer of this long conditional tree.

app/src/main/java/com/alphawallet/app/ui/WalletConnectActivity.java
```diff
@@ -102,6 +106,16 @@
 
     private ActionSheetDialog confirmationDialog;
     private ActionSheetDialog walletConnectDialog;
+    private ActionSheetDialog switchChainDialog;
+    private AddEthereumChainPrompt addEthereumChainPrompt;
+    private long switchChainDialogCallbackId = 1;
```
JamesSmartCell, line 111: Use a single ActionSheetDialog member - in this case use 'walletConnectDialog'. This way we can ensure there's not an existing dialog when it's called.

JamesSmartCell (changes requested):
Just a couple of small maintainability updates

app/src/main/java/com/alphawallet/app/ui/WalletConnectActivity.java
asif-finimble: Done. separated the show dialog code to another method.

app/src/main/java/com/alphawallet/app/ui/WalletConnectActivity.java
asif-finimble: Done.

asif-finimble:
Applied changes, rebased and force pushed.

JamesSmartCell (approved):

PR #2516, opened 2022-04-04
Show more NFT data

### LINKED ISSUE #2437: Show more collection-level data from OpenSea #3730
@oa-s would you help write up details and then assign to James?

https://github.com/AlphaWallet/alpha-wallet-ios/issues/3730#issuecomment-1034803414

justindg:
- Closes #2437 

Also reworked the activity flow as follows:

1. Load default token data from `NFTAsset` and `Token` objects:
- Image
- Token ID or Sequence ID
- Network Name
- Contract Address

2. Attempt to fetch asset from OpenSea API. If successful, load more data into the view:
- Asset Name
- Creator
- Token Standard/Schema
- Total Supply
- Number of Owners
- External Link
- Owner
- Last Sale
- Description
- Attributes (w/ Rarity %)

3. If (2) fails for any reason, load from asset metadata instead:
- Balance
- Asset Name
- External Link
- Description
- Attributes (No Rarity %)

image


app/src/main/java/com/alphawallet/app/viewmodel/TokenFunctionViewModel.java
```diff
@@ -718,75 +742,79 @@ private Intent completeTransferIntent(Context ctx, Token token, List<BigInteger>
...
+                Timber.e(e);
             }
-
-            trait.updateTraitRarity(count);
         }
 
-        traits.postValue(traitList);
+        if (!loadedFromApi)
+        {
+            NFTAsset a = token.getAssetForToken(tokenId);
```
JamesSmartCell, line 813: I think this is the wrong way round,

Since we are re-loading metadata, if the API call fails we should then fall back to the direct contract call, and if that fails then re-send the old data from the token onAssetMetadataError

JamesSmartCell (changes requested):

JamesSmartCell:
Approved, but github is not letting me add 'approved'.

PR #2524, opened 2022-04-04
Feature/Add-Swapzone.io-to-Dapps-List

vrogov:
Swapzone is a cryptocurrency exchange aggregator. We want to make our integration to your wallet, so that users will be able to exchange their crypto coins and tokens (Swapzone supports over 300 assets) with best rates and prices, custody-free, with no registration needed.


JamesSmartCell:
Hi @vrogov we can also add you to our main dapp page; we generally ask that you also list us in your mobile wallet list on your wallet explanation page.

Your website looks like a useful service, although it doesn't seem to pick up the ethereum wallet address if it's running inside the dapp browser. I can add your service in as-is but if you want it on the main dapp page (which is served from here: https://alphawallet.com/browser/) then you might want to make it a little more web3 friendly:

- detect you are running in a web3 host
- if so, only list EVM tokens (and contracts) as the transfer-from chains (eg Eth, Gnosis, Polygon, Klaytn, USDT etc)
- Auto-populate the token address and query user balance

PR #2525, opened 2022-04-04
Updated switch chain Action Sheet

### LINKED ISSUE #2504: Improve Switch Chain ActionSheet
When switching chains with WalletConnect (and DappBrowser), the prompt is still the placeholder text message which was put in a while back.

Check Zeplin for an official design for this action sheet configuration: If it exists implement it, if not we can flag this as an issue for Tomek.

asif-finimble:
Closes #2504


JamesSmartCell (changes requested):
Could you move the chain switch element into a new widget, and display it above the message widget?

We will probably want to re-use it for another purpose and it would be great to be able to do that without needing to write any new code.

asif-finimble:
@JamesSmartCell Moved the switch chain into widget.

JamesSmartCell:
> @JamesSmartCell Moved the switch chain into widget.

Great, also code now looks cleaner too. This is why I like widgets :)

PR #2527, opened 2022-04-04
Prevent drawing chart if there is insufficient data

### LINKED ISSUE #2520: Tapping this token in Wallet tab seem to hang the app somewhat reliably
Tapping this token in Wallet tab seem to freeze the app somewhat reliably

This is with: 0x0d590124d2faabbbdfa5561ccbf778914a50bcca

Final release candidate v3.53.3 available on 20220402

justindg:
Closes #2520

Token: dForce (USDX) - 0xeb269732ab75A6fD61Ea60b06fE994cD32a83549

History Chart is attempting to draw the chart with insufficient data, where it gets stuck in an infinite loop during `onDraw` method

The CoinGecko API returns insufficient data (just 1 data point) for the 1D and 1W chart.


JamesSmartCell (approved):
Nice

PR #2540, opened 2022-04-11
Refactor NFTAssetDetailActivity + Store OpenSea data to Realm

### LINKED ISSUE #2522: Store NFT data associated with OpenSeaAsset
Store the data into a RealmNFTAsset after a successful call to the OpenSea asset API.

(But remember to validate the data against the existing asset first; preserve 'good' data from the existing asset - if the API call from OpenSea returns a null don't overwrite existing data; it might be an issue with OpenSea - eg they don't currently handle contracts that generate their own metadata).

To test: watch this address on Rinkeby: 0xa20efc4b9537d27acfd052003e311f762620642d

Look for the 'MachineFi' NFTs; the metadata is produced directly from the contract, so OpenSea returns null for name, description and image.

justindg:
Closes #2522 

**NFTAssetDetailActivity**
Instead of dynamically adding entries, I made them static (but initially hidden) so that redundant info from OpenSea API and contract data will not interfere with each other. Previously, loading from one method will clear all the entries loaded from the other method.

**TokenFunctionViewModel**
Modified the flow a bit. Even if `OpenSeaAsset.isValid()` is `false` (no name, image or description), the value will still be posted to the Activity to check if there are any other meaningful data to display. After which, the data from the contract will still be loaded.

**NFTAsset**
When invoking `NFTAsset.updateFromRaw()`, the OpenSea data (via `NFTAsset.attachOpenSeaAssetData()`) from a successful API call will be checked for a valid name, image or description, and will overwrite null/empty values from the `assetMap`

### DIFF
--- a/app/src/main/java/com/alphawallet/app/widget/TokenInfoView.java
+++ b/app/src/main/java/com/alphawallet/app/widget/TokenInfoView.java
@@ -79,10 +79,6 @@ public class TokenInfoView extends LinearLayout
             TextView useView = getTextView(text.length());
             useView.setText(text);
         }
-        else
-        {
-            setVisibility(View.GONE);
-        }
     }
 
     public void setCurrencyValue(double v)
--- a/app/src/main/java/com/alphawallet/app/widget/TokenInfoView.java
+++ b/app/src/main/java/com/alphawallet/app/widget/TokenInfoView.java
@@ -4,7 +4,6 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.res.TypedArray;
 import android.net.Uri;
-import android.text.Spanned;
 import android.text.TextUtils;
 import android.util.AttributeSet;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/alphawallet/app/widget/TokenInfoView.java
```diff
@@ -17,44 +20,80 @@
...
     }
 
     public void setValue(String text)
     {
-        if (text.startsWith("http")) { setLink(); }
-        TextView useView = getTextView(text.length());
-        useView.setText(text);
+        if (!TextUtils.isEmpty(text))
+        {
+            if (TextUtils.isEmpty(valueStr))
```
JamesSmartCell, line 76: collapse this nested conditional into an && above

JamesSmartCell (changes requested):

PR #2546, opened 2022-04-12
Ignore case when filtering suggestions

### LINKED ISSUE #2543: Import Wallet screen's Seed tab should start with lowercase keyboard and perform case-insensitive check for presenting autocompletion
- [ ] Should start with lowercase (Look for the shift key in the keyboard in the screenshot)
- [ ] User input should be checked ignoring case when presenting autocompletion so if the user enters "ABA", or "Aba", "abandon" should be shown

justindg:
Closes #2543

Unfortunately, for this issue
> Should start with lowercase (Look for the shift key in the keyboard in the screenshot)

It varies from every device. We have already implemented the most common solution for this which is `inputType=textMultiline`, but some manufacturers still force capitalization for the first letter.

This PR alleviates this by:
1. Ignoring the case when filtering for suggestions
2. Converting the whole seed phrase to lower case whenever a suggestion is clicked OR before submitting the final seed

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/ImportSeedFragment.java
+++ b/app/src/main/java/com/alphawallet/app/ui/ImportSeedFragment.java
@@ -124,7 +124,7 @@ public class ImportSeedFragment extends ImportFragment implements OnSuggestionCl
     private void processSeed(View view)
     {
         this.seedPhrase.setError(null);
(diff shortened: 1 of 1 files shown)


justindg:
Forced lowercase locale to ENGLISH, as some locales will cause issues (eg. Turkish will display lowercase `i` without the dot)

JamesSmartCell (approved):

PR #2547, opened 2022-04-12
Optimize image display

justindg:
Since we are loading data from 3 sources, the image flickers when data rapidly changes. This PR attempts to smoothen out the process.

Current scenario (MachineFi token):
1. Image found in existing metadata:  **Load image**
2. OpenSea returns null image: **Display the fallback layout (token symbol)**
3. Contract fetch returns an image url: **Reload image**

With this PR:
1. Image found in existing metadata:  **Load image**
2. OpenSea returns null - **Do nothing since we already have an image displayed**
3. Contract fetch returns an image url: **Do nothing because it is the same as existing one**

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/NFTAssetDetailActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/NFTAssetDetailActivity.java
@@ -269,7 +269,7 @@ public class NFTAssetDetailActivity extends BaseActivity implements StandardFunc
     {
         if (asset != null)
         {
-            updateTokenImage(asset);
+            updateTokenImage(asset.getImage());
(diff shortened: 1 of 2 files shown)


JamesSmartCell:
This is a good stopgap, but I will write up the real solution in a new issue which will be a lot cleaner; essentially 
- completely decouple the UI elements from all the fetches.
- After each API/contract call, check to see if Realm needs updating and update if required.
- have a Realm listener which listens specifically for any change to the Asset metadata for this token
- Listener updates the view with the new metadata.

JamesSmartCell:
See #2548

JamesSmartCell (approved):

PR #2550, opened 2022-04-13
Added Ens Node not synced action sheet in SendPayment

### LINKED ISSUE #2544: Warn user about unsynced ENS send
If the user's phone time is not synced correctly, the wallet can fail to validate the user's ENS.

see this code in EnsResolver:

```
    public String resolve(String ensName)
    {
        String contractAddress = ensName;
        if (isValidEnsName(ensName, addressLength))
        {
            try
            {
                if (!isSynced()) //ensure node is synced
                {
                    throw new EnsResolutionException("Node is not currently synced");
                }
```

The ```isSynced()``` call checks the blocktime against the phone's current time, and if the phone's current time is not correct it can result in ENS not working (which is correct as it's picking up a possible attack).

If this error is thrown, the user should be warned that the node doesn't appear to be synced but that it may be the phone's time is not synced (the most likely cause).

```ENS Lookup Warning```
```AlphaWallet has detected a discrepancy between ethereum's timestamp and your device's current time, or possibly the blockchain connection is not synchronized, please check your phone's time is correct, or ignore this warning.```

The dialog should have 'Cancel' as the main action, and 'Ignore' for the secondary.

If Ignore is chosen, re-run the ENS call but disable the ```isSynced()``` check.

asif-finimble:
Closes #2544


JamesSmartCell (changes requested):
The warning should be an AWalletAlertDialog as it's not part of the normal flow and represents an fault. The idea with ActionSheet is it presents a choice to the user that is part of the normal flow, eg send transaction, sign, switch chains etc.

Other than that it looks good!

asif-finimble:
Ok. Will make the change.

JamesSmartCell:
@asif-finimble could you update your Android Studio with the ```AlphaWalletStyle.xml``` file in the repo and reformat the source.

asif-finimble:
Sorry for that James. I'll do it.

JamesSmartCell (approved):

PR #2557, opened 2022-04-14
Fix navbar alignment

justindg:
Noticed that we previously used "hacks" to align the navbar and the respective fragments. Added a permanent solution instead


JamesSmartCell:
> Noticed that we previously used "hacks" to align the navbar and the respective fragments. Added a permanent solution instead

Yes we did, a magic margin. Hopefully it won't ever go wrong again now.

JamesSmartCell (approved):

PR #2563, opened 2022-04-18
Add Network Display Widget for Action Sheet

### LINKED ISSUE #2554: Improve how we display Network in the ActionSheet
This is how it looks now:

Note that we indicate the network in the right corner of the Balance cell. Let's use another row. So below the "Balance" we will have one more row "Network". Then icon + Network name

Screenshot 2022-04-14 at 08 40 45

**Zeplin:**
https://zpl.io/dxD1PKL
https://zpl.io/awGoGOD (you can reuse this component, but ignore chevron)

justindg:
Closes #2554


justindg:
Screenshot_20220418-102103_AlphaWallet

JamesSmartCell:
looks good, I'll check how it looks on the Pixel with a Dapp transaction; we might need to respond to the smaller layout

JamesSmartCell (approved):

PR #2571, opened 2022-04-22
Change Add Token to Add.

### LINKED ISSUE #2236: Add wallet button content description shows "Add Token", it could be "Add Wallet"
image

image

vfbiby:
Fix #2236

### DIFF
--- a/app/src/main/res/menu/menu_add.xml
+++ b/app/src/main/res/menu/menu_add.xml
@@ -6,6 +6,6 @@
         android:id="@+id/action_add"
         android:icon="@drawable/ic_plus"
         android:orderInCategory="20"
-        android:title="@string/action_add_token"
+        android:title="@string/action_add"
         app:showAsAction="always" />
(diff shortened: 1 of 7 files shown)


JamesSmartCell (changes requested):
Hi @vfbiby thankyou for the submission.

To complete the change you'd need to change the string resources.

So, you've updated the string resource pointer, now you have to update all the actual string resources.

You can find these in the res/values/strings directory in the AndroidStudio project.

We have the default English, and also Chinese, French, Spanish, Vietnamese and Burmese. These other languages need to be updated too; you could try using Google Translate for each of them.

vfbiby:
action_add exists already, I deleted action_add_token, because there is no reference.

JamesSmartCell:
Ah ok, I see what you did previously.

Thanks for the update, transactions are great.

JamesSmartCell (approved):

PR #2577, opened 2022-04-26
Reworked numeric input sheet

justindg:
Noticed that `NumericInputBottomSheet` had a few bugs, and also it wasn't really a BottomSheet. Reworked the Numeric Input Sheet to allow a bit more scalability and reusability.

- Created two widgets, a standalone `QuantitySelector` widget and a `BottomSheetDialog` version `QuantitySelectorDialog` which utilizes the former.
- Updated ERC1155 transfer flow using the new widgets


JamesSmartCell (changes requested):
Just a couple of things:

- should the numeric input default to a '1' if it's a quantity selection (like in sending ERC1155)?
- I think if we have a multiple selection, we pick a number for one token, then click on another token, it should preserve that number. The numerical selection took quite a while to get exactly right when I did it first time, as you have to deal with quite a lot of edge conditions. To replicate this, watch jsbrown.eth and try selecting multiple tokens in the 'Chicken Giblets' collection.

It's a definite improvement though.

justindg:
> * should the numeric input default to a '1' if it's a quantity selection (like in sending ERC1155)?

Initially, I tried restricting the input using a min-max value but I wasn't happy with how intrusive it can be. But setting the default to 1 instead of 0 should be fine. I will test this.

> * I think if we have a multiple selection, we pick a number for one token, then click on another token, it should preserve that number. The numerical selection took quite a while to get exactly right when I did it first time, as you have to deal with quite a lot of edge conditions. To replicate this, watch jsbrown.eth and try selecting multiple tokens in the 'Chicken Giblets' collection.

How do we handle the scenario where tokens have different balances? 
Example - Token A with 4 assets and Token B with 3 assets. 
- Click Token A -> Select 4 assets -> Confirm
- Click Token B -> ???

JamesSmartCell:
> > * should the numeric input default to a '1' if it's a quantity selection (like in sending ERC1155)?
> 
> Initially, I tried restricting the input using a min-max value but I wasn't happy with how intrusive it can be. But setting the default to 1 instead of 0 should be fine. I will test this. [Update: branch updated, seems to work well.]
> 
> > * I think if we have a multiple selection, we pick a number for one token, then click on another token, it should preserve that number. The numerical selection took quite a while to get exactly right when I did it first time, as you have to deal with quite a lot of edge conditions. To replicate this, watch jsbrown.eth and try selecting multiple tokens in the 'Chicken Giblets' collection.
> 
> How do we handle the scenario where tokens have different balances? Example - Token A with 4 assets and Token B with 3 assets.
> 
> * Click Token A -> Select 4 assets -> Confirm
> * Click Token B -> ???

Click Token A->Select 4 assets
Click Token B ->Select 2 assets

Click Transfer to transfer 4 A + 2 B.

Currently the flow is:

Click Token A -> Select 4 assets -> Confirm
Click Token B -> Select 4 assets -> Confirm
Click Transfer.

What we want is:

Click Token A -> Select 4 assets
Click Token B -> Select 2 assets

Click Transfer.

If I try this, I get:

Click Token A -> Select 4 assets
Click Token B --> Token A blanked ...

JamesSmartCell (approved):

PR #2580, opened 2022-04-27
2517 send transaction timeout

### LINKED ISSUE #2517: SendTransaction Timeout
Sometimes Infura takes a long time to return the transaction, the socket times out but the dialog just hangs. Usually, the transaction has been sent and AW picks up the transaction as normal, in a minute or so.

Either: Add a 30 second timeout for Transaction Sending, which should be done in a non-duplicative way - sendTransaction is called from various places; however it always ends up in one of these methods:

```resendTransaction```
```createTransactionWithSig``` 
```create1559TransactionWithSig```

Or: See if hooking up the timeout from the AWHttpService is possible; it picks up other exceptions and relays them to the exception handler. You might have to experiment with some 'fake' exceptions or generate your own timeouts after a certain time.

### PR DESCRIPTION
**Pull request by asif-finimble**
Closes #2517

### DIFF
--- a/app/src/main/java/com/alphawallet/app/repository/TransactionRepository.java
+++ b/app/src/main/java/com/alphawallet/app/repository/TransactionRepository.java
@@ -91,7 +91,7 @@ public class TransactionRepository implements TransactionRepositoryType {
 
 	@Override
 	public Single<TransactionData> create1559TransactionWithSig(Wallet from, String toAddress, BigInteger subunitAmount, BigInteger gasLimit, BigInteger maxFeePerGas, BigInteger maxPriorityFee, long nonce, byte[] data, long chainId) {
-		final Web3j web3j = getWeb3jService(chainId, C.SEND_TRANSACTION_TIMEOUT);
+		final Web3j web3j = getWeb3jService(chainId);
 
 		TransactionData txData = new TransactionData();
 
@@ -121,7 +121,7 @@ public class TransactionRepository implements TransactionRepositoryType {
 
 	@Override
 	public Single<TransactionData> createTransactionWithSig(Wallet from, String toAddress, BigInteger subunitAmount, BigInteger gasPrice, BigInteger gasLimit, long nonce, byte[] data, long chainId) {
-		final Web3j web3j = getWeb3jService(chainId, C.SEND_TRANSACTION_TIMEOUT);
+		final Web3j web3j = getWeb3jService(chainId);
 		final BigInteger useGasPrice = gasPriceForNode(chainId, gasPrice);
 
 		TransactionData txData = new TransactionData();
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Review by JamesSmartCell — CHANGES_REQUESTED** [review]
Loos good, I have a few suggestions -

(I think) the timeout should be on the write, not the read, since we're sending a transaction. You can update this to 30 seconds universally, so no need for the extra parameter.

Second - can you catch the exception from the sendtransaction into a separate function that calls either the usual fail or the specific timeout fail - the conditional between two dialogs within the function is less flexible and more prone to breaking in the future.

**Comment by asif-finimble** [comment]
Ok. Will make the changes.   
Also the exception is thrown by viewModel via liveData. So all errors will be received in the ```txError()``` method. I'll separate out the different dialogs (timeout/error) into separate functions.

File: app/src/main/java/com/alphawallet/app/repository/TransactionRepository.java [hunk 1]
```diff
@@ -23,6 +29,7 @@
 import io.reactivex.Single;
 import io.reactivex.schedulers.Schedulers;
 import io.realm.Realm;
+import timber.log.Timber;
 
 public class TransactionRepository implements TransactionRepositoryType {
```
JamesSmartCell, line 35: Can the changes to this file be reverted? There doesn't appear to be any change here.

app/src/main/res/values/strings.xml
```diff
@@ -926,4 +926,6 @@
     <string name="unable_store_key">Unable to store key: %1$s</string>
     <string name="title_ens_lookup_warning">ENS Lookup Warning</string>
     <string name="message_ens_node_not_sync">AlphaWallet has detected a discrepancy between ethereum\'s timestamp and your device\'s current time, or possibly the blockchain connection is not synchronized, please check your phone\'s time is correct, or ignore this warning.</string>
+    <string name="message_transaction_timeout">The transaction has timed out. You can view the status in activity page.</string>
+    <string name="error_transaction_timeout">Transaction Timeout</string>
 </resources>
```
JamesSmartCell, line 931: What about the strings for the other languages?

JamesSmartCell (changes requested):

app/src/main/res/values/strings.xml
asif-finimble: Done

app/src/main/java/com/alphawallet/app/repository/TransactionRepository.java
asif-finimble: Done

app/src/main/res/values/strings.xml
JamesSmartCell: How about -es ?

app/src/main/java/com/alphawallet/app/repository/TransactionRepository.java
```diff
@@ -15,6 +16,11 @@
...
 import org.web3j.protocol.Web3j;
 import org.web3j.protocol.core.methods.response.EthSendTransaction;
+import org.web3j.protocol.core.methods.response.EthTransaction;
+import org.web3j.protocol.core.methods.response.TransactionReceipt;
+import org.web3j.rlp.RlpEncoder;
+import org.web3j.rlp.RlpList;
+import org.web3j.rlp.RlpType;
 import org.web3j.utils.Numeric;
 
 import java.math.BigInteger;
```
JamesSmartCell, line 26: Could you do a 'compare with origin/master' and sync to that?

JamesSmartCell (changes requested):

app/src/main/java/com/alphawallet/app/repository/TransactionRepository.java
asif-finimble: Sorry I missed -es.   
Compared and synced TransactionRepository with master

JamesSmartCell (approved):

PR #2602, opened 2022-05-09
Milkomeda chain support

micwallace:
This PR adds Milkomeda chain support. Milkomeda is a EVM compatible side-chain for the Cardano network. Milkomeda does not have a token of it's own, rather it's main asset is wrapped ADA, known as MilkADA. 

The PR includes the addition of Milkomeda chain definitions for main and tests nets, as well as Ticker service for mainnet.


JamesSmartCell:
Awesome :) How integrated is this to the main Cardano?

micwallace:
It's just a sidechain with a bridge for asset transfer to/from the main Cardano chain: 
https://cardano-bridge.milkomeda.com/bridge

They plan to bridge to Solana and Algorand too. I also saw something about turning it into an L2, but that's probably a long way away.

JamesSmartCell (approved):

PR #2671, opened 2022-06-22
2666 show test mode indicator

### LINKED ISSUE #2666: Mode Testnet Shows Balance Erroneously on Summary
Being on mode Testnet the app shows the summary balance of Mainnet. Also you can see that the app does not make a difference between Mainnet and Testnet as iOS does on the Wallets section. 

Android: 
image

iOS:
image

seabornlee:
fix #2666 

Also, add "Where are my Tokens?" on HomeActivity.


JamesSmartCell (changes requested):
This is really helpful (esp the 'switch to mainnet' button).

Just one small change, in dark mode it's very hard to see the 'where are my tokens?' text. Can you add theming so the blue used in dark mode here is the same as the blue highlight for the nav bar when enabled (eg Wallet, Activity etc).

Issue #2679, opened 2022-06-27
Redeploy DMZ service


hboon:
After #2678

foxgem:
what we need:

- how to build
- the deployment configuration, fox example, env variables?

weiwu-zhang:
Read the email "Redeploy DMZ service #2679"

This is a very early 2017 set up with James Sangalli so everything may look ancient to you.

hboon:
- [ ] A simple test that the important part is still working is:

```
curl https://aw.app/apple-app-site-association
```

Expect:

```
{
  "applinks": {
    "apps": [],
    "details": [
      {
        "appID": "LRAW5PL536.com.stormbird.alphawallet",
        "paths": [ "*" ]
      }
    ]
  }
}
```

- [ ] and that the latest has been deployed because this now gives a `404`:

```
curl https://aw.app/api/v1/verifyXMLDSig -d "" -i
```

Expected:

```
HTTP/2 404 
date: Thu, 30 Jun 2022 02:21:48 GMT
content-type: application/json
vary: Origin
vary: Access-Control-Request-Method
vary: Access-Control-Request-Headers

{"timestamp":"2022-06-30T02:21:48.642+0000","status":404,"error":"Not Found","message":"No message available","path":"/api/v1/verifyXMLDSig1"}
```

Not expected (`500` or `400`):

```
HTTP/2 500 
date: Thu, 30 Jun 2022 02:21:57 GMT
content-type: application/json

{"timestamp":"2022-06-30T02:21:57.966+0000","status":500,"error":"Internal Server Error","message":"Current request is not a multipart request","path":"/api/v1/verifyXMLDSig"}
```

abcfy2:
@hboon @weiwu-zhang Could I redeploy dmz service now ?

hboon:
@abcfy2 yes with the latest master. Meant to ping you. Thanks.

abcfy2:
@hboon Done. Please check it.

hboon:
@abcfy2 looks good. Thanks!

Issue #2687, opened 2022-06-30
java.lang.IllegalAccessError between realm and java 16


weiwu-zhang:
This seem to be Java related. 

There is this in `app/build.gradle`

    sourceCompatibility JavaVersion.VERSION_1_8

However, building with Java 16 does not produce any message such as incompatible JDK, but gives a very specific error.

1. Is this error caused by JDK compatibility?
2. If so, what versions are allowed?
3. Can we customise gradle to produce useful messages when incompatible jdk is used?

```
$ gradle build

Welcome to Gradle 7.3!

Here are the highlights of this release:
 - Easily declare new test suites in Java projects
 - Support for Java 17
 - Support for Scala 3

For more details see https://docs.gradle.org/7.3/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)
Checking the license for package Android SDK Build-Tools 32 in /opt/android/licenses
License for package Android SDK Build-Tools 32 accepted.
Preparing "Install Android SDK Build-Tools 32 (revision: 32.0.0)".
"Install Android SDK Build-Tools 32 (revision: 32.0.0)" ready.
Installing Android SDK Build-Tools 32 in /opt/android/build-tools/32.0.0
"Install Android SDK Build-Tools 32 (revision: 32.0.0)" complete.
"Install Android SDK Build-Tools 32 (revision: 32.0.0)" finished.
Checking the license for package Android SDK Platform 32 in /opt/android/licenses
License for package Android SDK Platform 32 accepted.
Preparing "Install Android SDK Platform 32 (revision: 1)".
"Install Android SDK Platform 32 (revision: 1)" ready.
Installing Android SDK Platform 32 in /opt/android/platforms/android-32
"Install Android SDK Platform 32 (revision: 1)" complete.
"Install Android SDK Platform 32 (revision: 1)" finished.

> Task :lib:compileJava
Note: /home/a/IdeaProjects/alpha-wallet-android/lib/src/main/java/com/alphawallet/token/tools/XMLDSigVerifier.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: /home/a/IdeaProjects/alpha-wallet-android/lib/src/main/java/com/alphawallet/token/tools/XMLDSigVerifier.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :app:mergeAnalyticsDebugResources
[...]

PR #2688, opened 2022-06-30
added build instructions 

weiwu-zhang:
This partially solves #2679 (the other part is solved by email)

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,0 +1,36 @@
+# Build instruction
+
+To build everything:
+
+```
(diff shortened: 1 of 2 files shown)


asif-finimble:
@weiwu-zhang @JamesSmartCell I opened #2691 by checking out from @weiwu-zhang 's branch (better-build-instructions) as per discussion in #2687 and added changes on top of that. As they are merged, I think this can be closed now.

JamesSmartCell:
Closed as this update is already passed

PR #2725, opened 2022-07-15
[Swap] Improve search filter on SelectTokenDialog

justindg:
Added a slight improvement to user experience when searching through the token list. With this change, the filtered list prioritizes entries that **starts with** the search filter, THEN all other entries containing the filter

For example, if your search filter is `ETH`, the token `Ethereum (ETH)` will most likely be on top of the list (if combined with changes from #2724).

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/widget/adapter/SelectTokenAdapter.java
+++ b/app/src/main/java/com/alphawallet/app/ui/widget/adapter/SelectTokenAdapter.java
@@ -95,11 +95,11 @@ public class SelectTokenAdapter extends RecyclerView.Adapter<SelectTokenAdapter.
         {
             if (data.name.toLowerCase(Locale.ENGLISH).contains(searchFilter.toLowerCase(Locale.ENGLISH)))
             {
-                filteredList.add(data);
+                if (!filteredList.contains(data)) filteredList.add(data);
             }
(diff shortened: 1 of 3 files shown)


seabornlee (changes requested):
Suggest eliminating duplications.
Pass `data.symbol` & `data.name` into a `match()` method.

justindg:
@seabornlee The TokenFilter class looks good, I'll check back to test if the filter works as desired.

justindg:
I modified the filter to achieve the desired order. To demonstrate further, here is an example of a simple search with keyword `ETH`

1) Double filter - ETH shows up first, which is more than likely what the user wants
292546601_600088614831823_3261977048091429420_n

2) Single filter:
293255745_564589131976857_3873232521465638742_n

JamesSmartCell (approved):

seabornlee:
> @seabornlee The TokenFilter class looks good, I'll check back to test if the filter works as desired.

Oh, sorry I didn't realize the order is critical. To express the intention, you can write a unit test other than add comments since unit tests can encourage refactoring.

seabornlee:
image

justindg:
> @justindg My result looks different than yours.

Have to update with latest master (the other PR included alphabetical sorting)

seabornlee:
> > @justindg My result looks different than yours.
> 
> Updated with latest master (the other PR included alphabetical sorting of the initial list)
Ok, thx!

seabornlee (approved):

PR #2726, opened 2022-07-15
Added options to delete WalletConnect history

### LINKED ISSUE #2722: Add option to delete WalletConnect history
Settings -> WalletConnect

Add a 'recycle bin' icon at top, when clicked have these options:

Delete Empty
Delete All
Cancel

If Delete Empty is called, delete all Historic Walletconnect sessions with no transaction or signing activity.
If Delete All is called - do that.

Don't delete any connection that's active.
Reset the view when deletion callback comes back (maybe try using the realm success callback like eg ```public void updateWalletItem(Wallet wallet, WalletItem item, Realm.Transaction.OnSuccess onSuccess)```

asif-finimble:
Closes #2722


JamesSmartCell (approved):
Nice work

PR #2727, opened 2022-07-17
Enhance Swap Token Sort

JamesSmartCell:
- Only allow positive balance tokens for source select.
- search by fiat value in source search
- search by alphabetical order in destination but keep native token first.


seabornlee:
I am adding unit tests for the sort.

JamesSmartCell:
> I am adding unit tests for the sort.

Note I simply copied how it appears in Uniswap, they always have base token first then the rest. For source token, if you want to change that it makes sense to only show source tokens the user has a balance in, listed in order of real value.

seabornlee:
For the target token, should the tokens with a positive balance show at the top of the result?

JamesSmartCell:
> @JamesSmartCell For the target token, should the tokens with a positive balance show at the top of the result?

I wasn't sure - I thought the target could be anything, but checking Uniswap and Quickswap again, they sort the same way for target, that is - fiat value followed by alphabetical. So let's just copy them, as user will be expecting this

PR #2729, opened 2022-07-18
[DMZ] Tweak message shown https://aw.app/wc is visited

hboon:
Occassionally certain dapp + WalletConnect usage combination results in the user loading https://aw.app/wc in iOS mobile Safari. This smoothens it a little, making it less confusing for users, but ideally users should never need this. Example to reproduce: https://github.com/AlphaWallet/alpha-wallet-ios/issues/4660#issuecomment-1139221167

Before|After
-|-
Screenshot 2022-07-18 at 3 22 39 PM|Screenshot 2022-07-18 at 3 21 55 PM


JamesSmartCell (approved):
good idea.

PR #2730, opened 2022-07-18
[DMZ] Remove /api/v1/verifyXMLDSig #2678

### LINKED ISSUE #2678: Stub/comment out sig verification endpoint in DMZ
This https://github.com/AlphaWallet/alpha-wallet-android/blob/abda84cb68c595caa5a53a0be42187c71829e04d/dmz/src/main/java/com/alphawallet/token/web/AppSiteController.java#L576

- [ ] Buzz Hu/Feng Yu to redeploy with #2679

hboon:
Closes #2678


hboon:
Need quick check to see it doesn't break other things (seems harmless to remove that 1 method)

PR #2738, opened 2022-07-21
2718 remove auto update feature

seabornlee:


JamesSmartCell:
@seabornlee I'm wondering if we can isolate this so it's only in the non-analytics build, that is, the build that goes to the play store wouldn't be asking for this permission.

That way we aren't removing a feature that would keep downloaded user's apps up to date

app/src/main/java/com/alphawallet/app/viewmodel/HomeViewModel.java
```diff
@@ -407,8 +337,6 @@ public void handleQRCode(Activity activity, String qrCode)
                     showActionSheet(activity, qrResult);
                     break;
                 case PAYMENT:
-                    showSend(activity, qrResult);
```
JamesSmartCell, line 410: Should this be removed? That is, it may be used.

JamesSmartCell (changes requested):
Can we have an intermediate version where users who downloaded the app from the website are told that a new version is available, but when clicking on the 'update' button are simply taken to the website?

app/src/main/java/com/alphawallet/app/viewmodel/HomeViewModel.java
seabornlee: Not removed, IDE suggests merging the 2 cases.

seabornlee:
> Can we have an intermediate version where users who downloaded the app from the website are told that a new version is available, but when clicking on the 'update' button are simply taken to the website?

Right now, the feature has no entry, and the code is not tested as well. So I prefer to remove it for now and rewrite it later.
I would like to research how to add this feature only for `noAnalytics` flavor.

JamesSmartCell:
@seabornlee What I meant was; remove the ability to upgrade (thus solving the base issue), but keep the (dismissable) update prompt in the settings page. It just doesn't do anything when you click on it. This may still be working correctly, I'm not sure.

app/src/main/java/com/alphawallet/app/viewmodel/HomeViewModel.java
JamesSmartCell: > Not removed, IDE suggests merging the 2 cases.

That'll learn me to not just look at the diff.

seabornlee:
> @seabornlee What I meant was; remove the ability to upgrade (thus solving the base issue), but keep the (dismissable) update prompt in the settings page. It just doesn't do anything when you click on it. This may still be working correctly, I'm not sure.

Ok, I will upgrade.

JamesSmartCell:
@seabornlee I will push this as is for now as we need this for the release, but let's test to see if download-installed users can be informed of an update in a new PR

JamesSmartCell (approved):

PR #2746, opened 2022-07-25
Exclude unsupported chains from swap settings list

justindg:
The LiFi API supports these networks, which are not available in AlphaWallet:

- OKXChain Mainnet (66)
- Harmony Mainnet Shard 0 (1666600000)
- Moonriver (1285)
- Moonbeam (1284)
- Celo Mainnet (42220)
- Fuse Mainnet (122)

This PR removes these from the list in Swap Settings


seabornlee (approved):
Awesome work!

JamesSmartCell (approved):

PR #2752, opened 2022-07-29
Update EIP712 signing code

JamesSmartCell:
Fixes an issue reported on the discord forum where the EIP712 v4 opensea confirmation signature was being calculated incorrectly.

Added unit tests with a sample Opensea issue reflecting the fix included.


hboon:
@JamesSmartCell do you have a summary of what this is about? Wondering if it is the reason why signing EIP712 for OpenSea + iOS is not working

PR #2754, opened 2022-07-30
Use Klaytn production node

### LINKED ISSUE #2749: Use Klaytn production RPC node for Cypress mainnet and Baobab testnet
```
if API basic auth credentials available
    Use it with Klaytn API Service (KAS)
else
    Use public/free node (like it is, now)
end
```

KAS RPC node URL is the same for both Cypress and Baobab:

```
https://node-api.klaytnapi.com/v1/klaytn
```

When using it, these HTTP headers have to be included:

```
[
    "Authorization": "Basic \(basicAuth)",
    "x-chain-id": "\(chainID)",
]
```

chainID = 8217 or 1001

Will send `basicAuth` privately to @JamesSmartCell.

---

Actually not sure what is the rate limit with KAS, but the public node that we used is from a shared pool with every user/app, so let's go with KAS and if necessary we can get a better plan later

JamesSmartCell:
Closes #2749


JamesSmartCell:
@seabornlee is the codecov showing a fail because there doesn't appear to be an associated unit test?

How would we notify codecov there was an associated test (if we added it)?

seabornlee (approved):

PR #2757, opened 2022-08-08
- Implemented non zero block testnet list in WalletsActivity when use…

keval-finimble:
Added horizontal list of blockchains in WalletsActivity to show case the non zero blocks when testnet is active.


seabornlee:
hi @keval-finimble , pls check the lint errors and Unit test job which is failed.

JamesSmartCell:
> hi @keval-finimble , pls check the lint errors and Unit test job which failed.

JamesSmartCell:
> > hi @keval-finimble , pls check the lint errors and Unit test job which failed.

re-open, accidentally closed

JamesSmartCell:
@keval-finimble can you check the source commit I just made and all the //TODO - JB comments. You can make the code quite a bit more simple by checking those.

Also you may want to simplify that error handler with the nested if statements, you can could check if you can chuck custom errors and handle those in their own methods.

PR #2779, opened 2022-08-25
2706 feature testnet balance

keval-finimble:
Implemented non zero block testnet horizontal list in WalletsActivity.


app/src/main/java/com/alphawallet/app/ui/widget/adapter/TestNetHorizontalListAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+    public void onBindViewHolder(@NonNull ViewHolder holder, int position)
+    {
+        Token token = tokens[position];
+        holder.tokenIcon.clearLoad();
+        try
+        {
+            String coinBalance = token.getStringBalanceForUI(4);
+            if (!TextUtils.isEmpty(coinBalance))
+            {
+                String symbol = token.getSymbol().substring(0, Math.min(token.getSymbol().length(), 5))
```
seabornlee, line 64: You can extract a method `getXXSymbol()` into `Token`, and write a unit test for it.

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TestNetHorizontalListAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+            // TODO - JB: use bindData(token.tokenInfo.chainId), remove import of assetDefinitionService
+            holder.tokenIcon.bindData(token.tokenInfo.chainId);
+            if (!token.isEthereum())
+            {
+                holder.tokenIcon.setChainIcon(token.tokenInfo.chainId); //Add in when we upgrade the design
+            }
+        }
+        catch (Exception ex)
+        {
+            System.out.println("Error" + ex);
```
seabornlee, line 78: `Timber.e(e);`

app/src/main/java/com/alphawallet/app/ui/widget/holder/WalletSummaryHolder.java
```diff
@@ -211,15 +256,25 @@ private Wallet fetchWallet(Wallet w)
...
             w.ENSAvatar = realmWallet.getENSAvatar();
-        }
+            if (w.tokens != null)
+            {
+                for (Token t : w.tokens)
+                {
+                    if (t.getWallet().equals(w.address))
+                    {
+                        testNetHorizontalListAdapter = new TestNetHorizontalListAdapter(w.tokens, context);
+                        recyclerView.setAdapter(testNetHorizontalListAdapter);
```
seabornlee, line 272: Shall we break the loop once matched?

seabornlee (changes requested):

app/src/main/java/com/alphawallet/app/ui/widget/holder/WalletSummaryHolder.java
keval-finimble: @seabornlee ,

I tried breaking the loop once the condition matched but it impacts the display time in the adapter. When we are not breaking a loop at that time the user experience is a bit smooth as compared to when we are breaking a loop. I can share you a video if needed. 

Please share your thoughts.

JamesSmartCell: Hi @keval-finimble there should not be any changes to WalletSummaryHolder. This file should be unchanged. We don't need to know the total testnet held - only find specific testnet balance on each wallet.

JamesSmartCell: Hi @keval-finimble can you move this handling out into the widget:

Send the w.tokens to your adapter and do all the handling there. Otherwise we have this several layers deep nested bracket; it makes the code more maintainable.

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TestNetHorizontalListAdapter.java
JamesSmartCell: @keval-finimble we actually have a standard method for displaying symbol in these conditions: checkout the source for```setupTextIcon``` in ```class TokenIcon```

app/src/main/java/com/alphawallet/app/ui/widget/adapter/WalletsSummaryAdapter.java
```diff
@@ -324,7 +334,32 @@ public int getDefaultWalletIndex()
...
 
-    public interface OnSetWalletDefaultListener {
+    public interface OnSetWalletDefaultListener
+    {
         void onSetDefault(Wallet wallet);
     }
+
+    public void setTokens(Token[] walletToken)
+    {
+        // This wont work.
```
JamesSmartCell, line 344: Remove comments

app/src/main/java/com/alphawallet/app/ui/widget/holder/WalletSummaryHolder.java
```diff
@@ -45,35 +53,62 @@ public class WalletSummaryHolder extends BinderViewHolder<Wallet> implements Vie
     private final ImageView defaultWalletIndicator;
     private final ImageView manageWalletBtn;
     private final UserAvatar walletIcon;
```
JamesSmartCell, line 55: Normalise member name (no underscore, this is only used in layouts).

app/src/main/java/com/alphawallet/app/ui/widget/holder/WalletSummaryHolder.java
```diff
@@ -45,35 +53,62 @@ public class WalletSummaryHolder extends BinderViewHolder<Wallet> implements Vie
...
-        walletClickLayout = findViewById(R.id.wallet_click_layer);
+        arrow_right = findViewById(R.id.container);
         wallet24hChange = findViewById(R.id.wallet_24h_change);
+        recyclerView = findViewById(R.id.horizontal_list);
         clickCallback = callback;
+        this.context = context;
+
         manageWalletLayout = findViewById(R.id.layout_manage_wallet);
         this.realm = realm;
     }
```
JamesSmartCell, line 91: Is the onError still used? If not it can be removed, along with the supporting Mutable

app/src/main/java/com/alphawallet/app/ui/widget/holder/WalletSummaryHolder.java
```diff
@@ -211,15 +248,19 @@ private Wallet fetchWallet(Wallet w)
...
                 .equalTo("address", w.address)
                 .findFirst();
-
         if (realmWallet != null)
         {
             w.balance = realmWallet.getBalance();
             w.ENSname = realmWallet.getENSName();
             w.name = realmWallet.getName();
             w.ENSAvatar = realmWallet.getENSAvatar();
-        }
```
JamesSmartCell, line 257: If the TestNetHorizontalAdapter is converted into a widget, deriving a LinearLayout there's no need to pass the context in. See eg SettingItemsView. Keep the TestNetHorizontalAdapter class but move it into a widget, which we can simply pass a Token[] array to and it displays the tokens. Then the context pass can be removed as it has its own context.

app/src/main/java/com/alphawallet/app/viewmodel/WalletsViewModel.java
```diff
@@ -57,18 +57,15 @@
...
-
     private final SetDefaultWalletInteract setDefaultWalletInteract;
     private final FetchWalletsInteract fetchWalletsInteract;
     private final GenericWalletInteract genericWalletInteract;
     private final FindDefaultNetworkInteract findDefaultNetworkInteract;
     private final KeyService keyService;
     private final ImportWalletRouter importWalletRouter;
     private final HomeRouter homeRouter;
-    private final TokensService tokensService;
+    public final TokensService tokensService;
```
JamesSmartCell, line 68: private TokensService

JamesSmartCell (changes requested):
Almost there @keval-finimble!

It might seem like a lot of changes but it's your first PR and there's a lot of nuances in this one (more than I realised when I assigned it).

app/src/main/java/com/alphawallet/app/ui/widget/holder/WalletSummaryHolder.java
```diff
@@ -211,15 +215,19 @@ private Wallet fetchWallet(Wallet w)
...
 
+            if (w.tokens != null)
+            {
+                TestNetHorizontalListAdapter testNetHorizontalListAdapter = new TestNetHorizontalListAdapter(w.tokens);
+                recyclerView.setAdapter(testNetHorizontalListAdapter);
+            }
+        }
         return w;
```
JamesSmartCell, line 231: this code should be moved into the widget, so the widget call should look like this:

```TokenWithBalanceView tokenBalanceView (maybe  rename to TokensBalanceView)```
...

```tokenBalanceView.populateTokens(tokens);```

where ```tokens``` is the Token[] array.

File: app/src/main/java/com/alphawallet/app/widget/TokenWithBalanceView.java [hunk 10]
```diff
@@ -0,0 +1,49 @@
...
+    TextView tokenPrice;
+
+    public TokenWithBalanceView(Context context, @Nullable AttributeSet attrs)
+    {
+        super(context, attrs);
+        inflate(context, R.layout.item_token_with_balance_view, this);
+        tokenIcon = findViewById(R.id.token_icon);
+        tokenPrice = findViewById(R.id.title_set_price);
+    }
+
```
JamesSmartCell, line 25: There should be a method in here that populates the internal adapter from a ```Token[]``` array.

JamesSmartCell (changes requested):
Not quite there yet.

The idea of having a widget is that it's self contained. I want to be able to send only a ```Token[]``` array to the widget and it handles everything.

JamesSmartCell:
@keval-finimble conflicts not resolved. Just use Github web interface.

JamesSmartCell (approved):

PR #2781, opened 2022-08-25
Allow users to purchase assets via Coinbase Pay

### LINKED ISSUE #2775: Allow purchase of crypto assets via Coinbase Pay
Allow users to purchase crypto assets via Coinbase Pay.

--

Endpoint: 
`https://pay.coinbase.com/buy/select-asset`

Parameters:
`appId` : `string` - Provided by Coinbase Pay team, check discord for valid AlphaWallet app ID.
`destinationWallets`  : `array` - See below.

Destination Wallets Parameters:
`address` : `string` -  Destination address where the purchased tokens will be sent.
`blockchains` : `string array` - List of enabled blockchains, available values*: `ethereum`, `solana`, `avalanche-c-chain` 
`assets` : `string array` - List of tokens that will be displayed to the user in the succeeding screen

* Seems only `ethereum` blockchain is relevant to AlphaWallet

Upon testing, there are two ways we could utilize this API.
1. Select from a list of tokens in the `ethereum` network:

```
https://pay.coinbase.com/buy/select-asset?appId=<APP_ID>&destinationWallets=[{"address": "<WALLET_ADDRESS>","blockchains":["ethereum"]}
```

2. Directly purchase a token; this will skip the token selection screen:

```
https://pay.coinbase.com/buy/select-asset?appId=<APP_ID>&destinationWallets=[{"address": "<WALLET_ADDRESS>","assets":["ETH"]}
```

Requirements to test this feature:
1. Verified Coinbase account - may include ID verification and linked payment info depending on the country of origin.

--

Sample flow:

https://user-images.githubusercontent.com/17334718/185843826-c267b688-f2c9-416b-b161-1bcc87d38522.mp4

justindg:
- Closes #2775


justindg:
I removed Coinbase Pay App ID for now, should we add it to our keys? From what I see, it is not really sensitive. If you use Coinbase Pay via Metamask, you can see the app id from the browser bar. But I'm leaving the decision to you @seabornlee @JamesSmartCell :)

app/src/main/java/com/alphawallet/app/util/CoinbasePayUtils.java
```diff
@@ -0,0 +1,32 @@
...
+import com.alphawallet.app.entity.coinbasepay.DestinationWallet;
+import com.google.gson.Gson;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class CoinbasePayUtils
+{
+
+    public static ArrayList<String> getAssetList(String tokenSymbol)
```
seabornlee, line 12: `Arrays.asList(...)` can do the same thing.

app/src/test/java/com/alphawallet/app/repository/CoinbasePayRepositoryTest.java
```diff
@@ -0,0 +1,44 @@
...
+        String expectedEncodedUri = "https://pay.coinbase.com/buy/select-asset?appId=" + CoinbasePayRepository.APP_ID + "&destinationWallets=%5B%7B%22address%22%3A%220x1234%22%2C%22assets%22%3A%5B%22ETH%22%5D%7D%5D";
+        assertThat(actualUri, equalTo(expectedEncodedUri));
+
+        try
+        {
+            String decodedUri = URLDecoder.decode(actualUri, "UTF-8");
+            String expectedDecodedUri = "https://pay.coinbase.com/buy/select-asset?appId=" + CoinbasePayRepository.APP_ID + "&destinationWallets=[{\"address\":\"0x1234\",\"assets\":[\"ETH\"]}]";
+            assertThat(decodedUri, equalTo(expectedDecodedUri));
+        }
+        catch (UnsupportedEncodingException e)
```
seabornlee, line 39: No need to try catch, throw it directly.

seabornlee (changes requested):

app/src/main/java/com/alphawallet/app/util/CoinbasePayUtils.java
justindg: Thanks. Turns out this was an unused method anyway. Removed :)

JamesSmartCell:
Hi @justindg do you need the salting tool to create the key?

justindg:
> Hi @justindg do you need the salting tool to create the key?

Hi James, if we decide to include it in keys, then I think yes. How do I access it?

JamesSmartCell:
> > Hi @justindg do you need the salting tool to create the key?
> 
> Hi James, if we decide to include it in keys, then I think yes. How do I access it?

I haven't added it to github yet. I'll check it in. However you'll probably want the command line tool, it's easy to use just > obfuscator <key to be salted>. Actually it's C++, not sure how to make a mac executable, but it should be straightforward using gcc.

JamesSmartCell (approved):

PR #2813, opened 2022-09-06
Revise i18n and enable Indonesia language & currency

seabornlee:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=LP9dsel)


JamesSmartCell (changes requested):
Use the Indonesian flag from the Zeplin resources

JamesSmartCell (approved):

PR #2816, opened 2022-09-07
Added boolean to TokensAdapter's constructor to handle the Where ar…

keval-finimble:
- Handled "Where are my tokens" visibility to resolve ticket 2799.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokensAdapter.java
+++ b/app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokensAdapter.java
@@ -59,7 +59,7 @@ public class TokensAdapter extends RecyclerView.Adapter<BinderViewHolder> {
     private boolean debugView = false;
 
     private boolean gridFlag;
-    private boolean testNetWhereAreMyTokensViewFlag;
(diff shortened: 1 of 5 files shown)


JamesSmartCell (changes requested):
I would rename the variable to 'boolean showTestNetTips;' and use a separate method to set it true, otherwise it should be false; rather than adding a boolean to the constructor; since it's only used in one place.

This reduces the number of files touched.

JamesSmartCell:
@keval-finimble bump. Any progress?

JamesSmartCell (changes requested):
Formatting

app/src/main/java/com/alphawallet/app/ui/widget/adapter/TokensAdapter.java
```diff
@@ -477,6 +478,11 @@ public void setFilterType(TokenFilter filterType)
         filterAdapterItems();
     }
```
JamesSmartCell, line 480: Fix formatting

JamesSmartCell (changes requested):
Fix formatting

JamesSmartCell (approved):

PR #2825, opened 2022-09-15
2758 feature customisation system

keval-finimble:
- Handled CustomSettings from json file

### DIFF
--- a/app/src/main/java/com/alphawallet/app/App.java
+++ b/app/src/main/java/com/alphawallet/app/App.java
@@ -22,10 +22,13 @@ import timber.log.Timber;
 public class App extends Application
 {
 
+    private static Context mContext;
+
     @Override
     public void onCreate()
     {
         super.onCreate();
+        mContext = getApplicationContext();
         Realm.init(this);
 
         if (BuildConfig.DEBUG)
@@ -67,5 +70,9 @@ public class App extends Application
 //		LockManager<CustomPinActivity> lockManager = LockManager.getInstance();
 //		lockManager.enableAppLock(this, CustomPinActivity.class);
 //		lockManager.getAppLock().setShouldShowForgot(false);
+    }
+
+    public static Context getContext() {
+        return mContext;
     }
 }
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/alphawallet/app/App.java
```diff
@@ -68,4 +71,8 @@ else if (mode == UiModeManager.MODE_NIGHT_NO)
 //		lockManager.enableAppLock(this, CustomPinActivity.class);
 //		lockManager.getAppLock().setShouldShowForgot(false);
     }
+
+    public static Context getContext() {
+        return mContext;
+    }
 }
```
JamesSmartCell, line 78: This is not production code. It will crash during use.

app/src/main/java/com/alphawallet/app/ui/NewSettingsFragment.java
```diff
@@ -334,8 +349,19 @@ private void addSettingsToLayout()
...
+            JSONObject customSettingsJsonObject = new JSONObject(lockedChains);
+            JSONArray chainsArray = customSettingsJsonObject.getJSONArray("locked_chains");
+            if (chainsArray.length() == 0)
+            {
+                systemSettingsLayout.addView(selectNetworksSetting, systemIndex++);
+            }
+        }catch (JSONException err){
+            err.printStackTrace();
+        }
+
```
JamesSmartCell, line 365: FORMATTING. USE THE ALPHAWALLETCONFIG.XML

Also, chains handling code should be common in the config file. You may need to refactor the custom view settings file so it acts as an instance of the viewmodel class that is using it, that way it can use the context of the hosting class, and be sure the context pointer of the custoviewsettings class will be destroyed before the host context is.

app/src/main/java/com/alphawallet/app/util/Utils.java
```diff
@@ -314,6 +324,21 @@ public static String loadJSONFromAsset(Context context, String fileName) {
...
+            JsonElement json=new Gson().fromJson(reader,JsonElement.class);
+            returnString = json.toString();
+        }catch (IOException ex){
+            ex.printStackTrace();
+            return null;
+        }
+        return returnString;
+    }
+
+
```
JamesSmartCell, line 340: FORMATTING

JamesSmartCell (changes requested):
It looks like you're going down the right path, just need to refactor because the code will result in crashes under certain conditions.

PR #2849, opened 2022-09-27
2758 feature customisation

keval-finimble:
- Handled CustomSettings through json file.

### DIFF
--- a/app/src/main/assets/custom_view_settings.json
+++ b/app/src/main/assets/custom_view_settings.json
@@ -1,9 +1,15 @@
 {
-  "exclusive_chains" :[
+  "exclusive_chains" :[{"chain" : 1}],
+  "locked_chains" : [{"chain" : 1}, {"chain" : 4}, {"chain" : 10}],
+  "locked_tokens" : [
     {
-      "chain" : 1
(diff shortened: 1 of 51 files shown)


JamesSmartCell (changes requested):
1. Code Formatting. Many places formatting is bad. Use the Alphawallet config file.
2. Please change the name of the service to 'CustomSettings'.
3. In many places you add the service to function calls within the viewModel which already have access to the service; please fix these.
4. In some places, calls from the activity pass the service back into the viewModel. This isn't necessary.
5. Please move as much processing as possible to the CustomSettings class. EG in NewSettingsFragment/ViewModel there's JSON handling code. This should be in the CustomSettings.
6. Code Formatting

JamesSmartCell:
@keval-finimble this PR uses git merge instead of git rebase master. 

Previously I stated don't use git merge as it makes it impossible to see the code change.

JamesSmartCell:
@keval-finimble please revert the git merge / restart the PR from current baseline master.

When doing a sync use > git rebase origin/master

If there are merge conflicts you'll have to go through each one and resolve it by hand.

PR #2851, opened 2022-09-28
Display OpenSea NFT Rarity Data if available

### LINKED ISSUE #2843: Implement displaying NFT rarity (not attribute rarity) from  OpenSea API
Keywords to look for: [OpenSea OpenRarity]

If so, let's figure out how to get it in

justindg:
- Closes #2843 

image

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/opensea/OpenSeaAsset.java
+++ b/app/src/main/java/com/alphawallet/app/entity/opensea/OpenSeaAsset.java
@@ -90,7 +90,7 @@ public class OpenSeaAsset
 
     @SerializedName("rarity_data")
     @Expose
-    public RarityData rarityData;
+    public Rarity rarity;
 
     public static class Collection
     {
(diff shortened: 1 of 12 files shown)


seabornlee:
Nice job! Only two minor feedback:
1. Name `RarityData` can be `Rarity`, according to <Clean Code> naming, a suffix such as Data/Info/Record is redundant.
2. i18n, you can use Google Translator, even not so accurate, it's better than nothing.

seabornlee (changes requested):
Nice job! Only two minor feedback:

1. Name RarityData can be Rarity, according to naming, a suffix such as Data/Info/Record is redundant.
2. i18n, you can use Google Translator, even not so accurate, it's better than nothing.

JamesSmartCell:
> Nice job! Only two minor feedback:
> 
> 1. Name RarityData can be Rarity, according to naming, a suffix such as Data/Info/Record is redundant.
> 2. i18n, you can use Google Translator, even not so accurate, it's better than nothing.

On the translation issue - we should take a leaf from Seaborn's book here - I've also started using Google translate for all the regions. It takes a bit longer to do, for future maybe we can write a small script that produces the translations for each language? I think you have to have a google developer license to use the translate API, it used to work for free.

Second thing is storage - should we store these extra values?

JamesSmartCell (approved):

seabornlee (approved):

PR #2861, opened 2022-10-03
Add JSON Customisation feature

keval-finimble:
- Did all the customization changes in this branch to create the new PR


JamesSmartCell:
The following points need to be resolved for the PR to be merged:

- [ ] Remove formatting changes to all code that doesn't need to be touched (this means that the code you write should be formatted correctly but don't change formatting of code you don't touch). The reason for this is it makes the PR very difficult to assess, because of large apparent code changes.
- [ ] Review the core file, CustomSettings and de-duplicate. Currently the code can't be merged - call a universal function first that checks if the JSON needs to be loaded due to memory scavenging, if it needs to be re-loaded then reload. Then proceed with returning the value asked for.
- [ ] Resolve merge conflicts WITHOUT using git merge. git rebase or manually editing the files is fine.

PR #2881, opened 2022-10-24
2864 refactor dappbrowser url bar to widget

seabornlee:


JamesSmartCell (commented):
Excellent! Great to see all that moved out of DappBrowserFragment. Will do testing and merge as priority.

JamesSmartCell (approved):

PR #2895, opened 2022-10-31
Analytics Update

justindg:
Reworked analytics service.  Tracking is now done through viewmodels. 

`Analytics.java` will contain all the property keys and enums.

`BaseViewModel` has new streamlined track methods:
- `track(Analytics.Navigation event)`  
- `track(Analytics.Navigation event, AnalyticsProperties props)`  
- `track(Analytics.Action event)`  
- `track(Analytics.Action event, AnalyticsProperties props)` 
- `trackError(Analytics.Error source, String message)`

Setting up your `ViewModel` to track events:
1. Make sure your ViewModel extends `BaseViewModel`.
2. Include `AnalyticsServiceType analyticsService` in the viewmodel constructor.
3. Invoke `setAnalyticsService(analyticsService)` at the end of constructor.

Tracking Navigation Events:
Navigation events are used mostly to track which screen the user is in. The preferred way is to call `viewModel.track()` during `onResume()`.

Tracking Action Events:
Action Events are used to track significant user activity i.e. importing wallet, buying with Coinbase Pay, etc. Call `viewModel.track()` within the desired method in the `Activity` or `ViewModel`.

Tracking Error Events:
Error Events will help us determine which errors the users frequently encounter. The preferred way is to call `viewModel.trackError()` within `try-catch` scenarios or on generic error events like `viewModel.onError()`.

Adding more data to your event:
```
AnalyticsProperties props = new AnalyticsProperties();
props.put(<KEY>, <VALUE>);
viewModel.track(<EVENT>, props);
```

Sample result:
image

image


app/src/main/java/com/alphawallet/app/viewmodel/BaseViewModel.java
```diff
@@ -1,109 +1,176 @@
...
+        else
+        {
+            String message = throwable.getMessage();
+            if (TextUtils.isEmpty(message))
+            {
+                error.postValue(new ErrorEnvelope(C.ErrorCode.UNKNOWN, null, throwable));
+            }
+            else
+            {
+                error.postValue(new ErrorEnvelope(C.ErrorCode.UNKNOWN, message, throwable));
```
seabornlee, line 101: Move `error.postValue(...` out of `if else`. Get 2 benefits:
1. Eliminate duplication
2. Easier to understand what is `if else` for

app/src/main/java/com/alphawallet/app/viewmodel/ImportWalletViewModel.java
```diff
@@ -40,43 +41,61 @@ public class ImportWalletViewModel extends BaseViewModel implements OnSetWatchWa
...
-                .subscribe(this::onWallet, this::onError).isDisposed();
+                .subscribe(w -> onWallet(w, ImportWalletType.PRIVATE_KEY), this::onError).isDisposed();
+    }
+
+    public void onSeed(String walletAddress, KeyService.AuthenticationLevel level)
+    {
+        if (walletAddress == null)
+        {
+            progress.postValue(false);
+            Timber.e("ERROR");
```
seabornlee, line 87: ERROR better be more detailed. Such as `walletAddress is null`. So that it's easier to locate in log file.

seabornlee (changes requested):
Legend work!

app/src/main/java/com/alphawallet/app/viewmodel/BaseViewModel.java
justindg: I think this is very old code, but got highlighted when I formatted it. But awesome catch nonetheless! :)

app/src/main/java/com/alphawallet/app/viewmodel/ImportWalletViewModel.java
justindg: I think this is very old code, but got highlighted when I formatted it. But awesome catch nonetheless! :)

justindg: Done!

app/src/main/java/com/alphawallet/app/viewmodel/BaseViewModel.java
justindg: Done!

seabornlee (approved):

JamesSmartCell (approved):

PR #2899, opened 2022-11-02
Fix for NPE encountered

JamesSmartCell:
Fixes namehash NPE - exception should be passed down the stackframe.

### DIFF
--- a/app/src/test/java/com/alphawallet/app/ENSTest.java
+++ b/app/src/test/java/com/alphawallet/app/ENSTest.java
@@ -5,13 +5,13 @@ import static com.alphawallet.ethereum.EthereumNetworkBase.MAINNET_ID;
 import static com.alphawallet.ethereum.EthereumNetworkBase.RINKEBY_ID;
 import static com.alphawallet.ethereum.EthereumNetworkBase.ROPSTEN_ID;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
 
 import com.alphawallet.app.service.AWHttpService;
 import com.alphawallet.app.util.ens.AWEnsResolver;
 import com.alphawallet.app.web3j.ens.Contracts;
 import com.alphawallet.app.web3j.ens.EnsResolutionException;
 
+import org.junit.Assert;
 import org.junit.Before;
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/alphawallet/app/ENSTest.java
```diff
@@ -103,16 +105,46 @@ public void testReverseResolve() throws Exception {
...
+
+            assertEquals(
+                    nameHash("eth"),
+                    ("0x93cdeb708b7545dc668eb9280176169d1c33cfd8ed6f04690a0bcc88a93fc4ae"));
+
+            assertEquals(
+                    nameHash("foo.eth"),
+                    ("0xde9b09fd7c5f901e23a3f19fecc54828e9c848539801e86591bd9801b019f84f"));
+        }
+        catch (Exception e)
```
seabornlee, line 122: No need to catch exception in test, can just throw it out, the test still fail.

app/src/test/java/com/alphawallet/app/ENSTest.java
```diff
@@ -103,16 +105,46 @@ public void testReverseResolve() throws Exception {
...
+        {
+            System.out.println("Correctly got the exception");
+            e.printStackTrace();
+        }
+        catch (Exception e)
+        {
+            //should be ENSResolutionException
+            System.out.println("Should be EnsResoltionException");
+            fail();
+        }
```
seabornlee, line 148: In JUnit 4, can test Exception this way:
```
@Rule
public ExpectedException exceptionRule = ExpectedException.none();

@Test
public void whenExceptionThrown_thenRuleIsApplied() {
    exceptionRule.expect(NumberFormatException.class);
    exceptionRule.expectMessage("For input string");
    Integer.parseInt("1a");
}
```

seabornlee (approved):

app/src/test/java/com/alphawallet/app/ENSTest.java
JamesSmartCell: The source said the ExpectedException syntax is deprecated, use 'Assert.assertThrows' syntax instead.

PR #2902, opened 2022-11-03
Fix TapJacking

seabornlee:
https://developer.android.com/topic/security/risks/tapjacking

Tested with TapJacker, Android 7:
https://github.com/seabornlee/tapjacker


justindg (approved):
Great catch!

JamesSmartCell:
@seabornlee will this break the build as the 'add custom chain' method is expected to return a boolean in case the RPC was not a valid URL (ie contains hijack code) ?

seabornlee:
@JamesSmartCell 
It's weird, it is still return void on my master branch after pulling, but I can see it returns boolean on GitHub. Any suggestions?

PR #2905, opened 2022-11-03
Provide JSON Customisation file

keval-finimble:


seabornlee:
@JamesSmartCell How about inject `CustomSettings` directly into Activity/Fragment?
Right now it's in ViewModel but all invocations are `viewModel.getCustomSettings().xxx()`, it's kind of `Man in the middle` smell. https://refactoring.guru/remove-middle-man

keval-finimble:
> @JamesSmartCell How about inject `CustomSettings` directly into Activity/Fragment? Right now it's in ViewModel but all invocations are `viewModel.getCustomSettings().xxx()`, it's kind of `Man in the middle` smell. https://refactoring.guru/remove-middle-man

JamesSmartCell:
> @JamesSmartCell How about inject `CustomSettings` directly into Activity/Fragment? Right now it's in ViewModel but all invocations are `viewModel.getCustomSettings().xxx()`, it's kind of `Man in the middle` smell. https://refactoring.guru/remove-middle-man

So inject directly into the Activity? So it would be like a second viewModel where it's used. Can we do that? If so, in this case the code would look cleaner.

@keval-finimble can you see if it's possible to do this? It should be --- unless there's an injection limitation

JamesSmartCell:
Also @seabornlee @keval-finimble please do not merge master in here! If you want to bring this up to date with the latest master please use rebase not merge.

JamesSmartCell:
@keval-finimble Also the cache mappings haven't been wired up - let me know if you need any clarification on what they're supposed to do.

PR #2909, opened 2022-11-04
fix malicious NFT make app crash

seabornlee:
- Fix image height
- Catch NumberFormatException when bgColor is invalid

### DIFF
--- a/app/src/main/java/com/alphawallet/app/widget/NFTImageView.java
+++ b/app/src/main/java/com/alphawallet/app/widget/NFTImageView.java
@@ -140,7 +140,7 @@ public class NFTImageView extends RelativeLayout
     {
         if (!Utils.stillAvailable(getContext())) return;
 
-        setWebViewHeight(getLayoutParams().height);
+        setWebViewHeight((int)getLayoutParams().width);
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/alphawallet/app/widget/NFTImageView.java
```diff
@@ -138,21 +140,22 @@ private void loadImage(String url, String backgroundColor, int corners) throws I
     {
         if (!Utils.stillAvailable(getContext())) return;
 
-        setWebViewHeight((int)getLayoutParams().width);
+        setWebViewHeight(getLayoutParams().height);
```
justindg, line 143: I remember there used to be NFTs which that have extremely long and they covered the entire screen, and there were NFTs which were extremely wide as well. The conclusion was for me to use the code above to force it into a square, as a compromise. However I am not sure if that is still relevant because it has been a while.

seabornlee: Thx for sharing this, I didn't find out what's the intention. I will revert it back.

JamesSmartCell (approved):

PR #2911, opened 2022-11-04
fix malicious dapp make app crash

seabornlee:
- Fix potential NPE
- Fix crash when sign invalid message


justindg (approved):
Good find!

JamesSmartCell (approved):

PR #2937, opened 2022-11-11
- Created new branch to merge the latest code for CustomisationSettin…

keval-finimble:


seabornlee:
Check all method references, make it private if possible.
image

slumericanB (approved):

JamesSmartCell:
@keval-finimble I'm not sure why we've gone back to static - is there any reason for that, and not going via the path @seabornlee suggested to try to inject the CustomSettings into each activity that uses it?

As it stands this approach will fail - because Android OS memory scavenging may (and will at some point) blank the mappings and the context you set up at App creation.

The previous approach was very close to working! Hopefully you still have the code.

The direction I would have gone with that approach was:

- see if it's possible to inject directly into the activity - if not, keep in viewmodel, if it is, use that method.
- see that using the double ended queue for the JSON settings is maybe not optimal (that was my bad!) and use mappings for each entry. The first two functions you can refactor to be much cleaner are the ```tokenCanBeDisplayed```, ```alwaysShow``` and ```isLockedToken```.

You'll need a 2-dimensional mapping like this:

```private final LongSparseArray<Map<String, TokenInfo>> lockedTokens = new LongSparseArray<>();```

In terms of psudocode you could do something like this:

```tokenCanBeDisplayed(...)
{
  restoreMappings(); // function that checks we have loaded the mappings and if not, re-load them
  return token.type == ContractType.ETHEREUM || token.isEnabled || isLockedToken(token.getChain(), token.getAddress());
}
```

```
    public static Boolean isLockedToken(long chainId, String contractAddress)
    {
        restoreMappings();
        Map<String, TokenInfo> mapping = lockedTokens.get(chainId);
        return mapping.containsKey(contractAddress.toLowerCase(Locale.ROOT));
    }
```

Make sure you use ```ConcurrentHashMap``` for the inner map for lockedTokens.

The restoreMappings(); function will just re-load the mappings if the OS has scavenged, or if it's the first time being used. You can use this function for all loading, as you don't need to load the data until it's needed the first time.

Locked and exclusive chains are fine to use the ```ConcurrentLinkedQueue``` of Longs and just use the contains method to check if those have been specified.

Hopefully you can restore the previous work, I don't think you can make this method work as you need ```context``` to re-load the data, and the entry/re-entry point for the app might not be ```HomeActivity```. The user might have paged the app out, used something memory hungry like playing a game or geocaching then paged back to AW, which might be one or two activities down from the HomeActivity, so if you trigger a reload check in HomeActivity's ```onResume()``` there it wouldn't fire and you'd be working with blank data (or NPE if you didn't add null guards or exception handling).

Does it make sense?

PR #2945, opened 2022-11-14
CustomSettings Implementation

keval-finimble:
* Enhanced the cache functionality by changing the return type of lockedTokens in CustomSettings.
* Added other fields in the json and call them directly from the json file in CustomSettings.


app/src/main/java/com/alphawallet/app/service/TokensService.java
```diff
@@ -105,12 +104,16 @@
     private Disposable openSeaQueryDisposable;
 
     private static boolean done = false;
+    CustomSettings customSettings;
```
JamesSmartCell, line 108: Why not import the CustomSettings directly from DI? This way you don't need context

JamesSmartCell (changes requested):
You need to try to inject the singleton class rather than simply initialising a local instance of the class, in all cases where CustomSettings is used.

If the inject doesn't work then revert to using via viewModel.

PR #2947, opened 2022-11-15
CustomSettings Implementation

keval-finimble:
* Enhanced the cache functionality by changing the return type of lockedTokens in CustomSettings.
* Added other fields in the json and call them directly from the json file in CustomSettings.


JamesSmartCell (changes requested):
1. Remove the static in CustomSettings and ensure CustomSettings is only accessed via the ViewModel. 
2. Refactor CustomSettings as previously directed.

PR #2952, opened 2022-11-17
Removed PHI network and associated items from dapps list

justindg:
Removed PHI network and associated items from dapps list.


seabornlee (approved):

Phinetwork:
We help you find a bug yesterday that's burning tokens and now this. there is special
Place in hell for sell outs like you

PR #2961, opened 2022-11-22
Normalize WalletConnect V2 Sign Method Dialog UI

justindg:
image


JamesSmartCell (changes requested):
Hi @justindg this issue was specifically for Transaction signing view not message signing; there was a large signing refactor done specifically for the side-by-side issue (Transaction Signing vs Message Signing). I should have been more specific I figured 'Transaction Dialog' would be sufficiently clear.

We can probably use some of the changes to the address view and others though.

Can you archive this and re-do the issue for the transaction signing alone? I'll show images next time so it's less ambiguous.

PR #2984, opened 2022-11-29
- Integration Test for RPC nodes

keval-finimble:


JamesSmartCell (changes requested):
Good start - can you also add to the test to verify that each network doesn't display a :( face. I am not sure the best way to do this; a good way could be to sense the "IC" value in the environment variables (I think it is "IC") and if sensed then if any network comes up with a :( switch on some text (or symbol) at the top of the page. The integration test would sense that text. 

To do this correctly you'd also need to display some different text (or symbol) once all networks have completed their check.

This way you can timeout, complete if the 'all network success' symbol/text comes up and mark fail if the 'one network has failed' symbol/text becomes visible.

This symbol/text should only appear if the app senses that it is running an integration test. I think the "IC" symbol will appear in the environment vars if it is (check the script).

seabornlee:
> Good start - can you also add to the test to verify that each network doesn't display a :( face. I am not sure the best way to do this; a good way could be to sense the "IC" value in the environment variables (I think it is "IC") and if sensed then if any network comes up with a :( switch on some text (or symbol) at the top of the page. The integration test would sense that text.
> 
> To do this correctly you'd also need to display some different text (or symbol) once all networks have completed their check.
> 
> This way you can timeout, complete if the 'all network success' symbol/text comes up and mark fail if the 'one network has failed' symbol/text becomes visible.
> 
> This symbol/text should only appear if the app senses that it is running an integration test. I think the "IC" symbol will appear in the environment vars if it is (check the script).

Do you mean "CI" value in environment variables?

JamesSmartCell:
> > Good start - can you also add to the test to verify that each network doesn't display a :( face. I am not sure the best way to do this; a good way could be to sense the "IC" value in the environment variables (I think it is "IC") and if sensed then if any network comes up with a :( switch on some text (or symbol) at the top of the page. The integration test would sense that text.
> > To do this correctly you'd also need to display some different text (or symbol) once all networks have completed their check.
> > This way you can timeout, complete if the 'all network success' symbol/text comes up and mark fail if the 'one network has failed' symbol/text becomes visible.
> > This symbol/text should only appear if the app senses that it is running an integration test. I think the "IC" symbol will appear in the environment vars if it is (check the script).
> 
> Do you mean "CI" value in environment variables?

That's exactly what I mean :)

seabornlee:
Hi @JamesSmartCell, the test is currently failing and I’m unsure whether it’s due to issues with the test code or if some of the networks are not responding. Could you please verify the network status on the test device?

JamesSmartCell:
> Hi @JamesSmartCell, the test is currently failing and I’m unsure whether it’s due to issues with the test code or if some of the networks are not responding. Could you please verify the network status on the test device?

Yes! Artis has been switched off (we checked with the Artis team), so it needs to be removed. Then the test should work. I can see it failing instantly because Artis isn't responding.

JamesSmartCell:
> > Hi @JamesSmartCell, the test is currently failing and I’m unsure whether it’s due to issues with the test code or if some of the networks are not responding. Could you please verify the network status on the test device?
> 
> Yes! Artis has been switched off (we checked with the Artis team), so it needs to be removed. Then the test should work. I can see it failing instantly because Artis isn't responding.



> Hi @JamesSmartCell, the test is currently failing and I’m unsure whether it’s due to issues with the test code or if some of the networks are not responding. Could you please verify the network status on the test device?



> Hi @JamesSmartCell, the test is currently failing and I’m unsure whether it’s due to issues with the test code or if some of the networks are not responding. Could you please verify the network status on the test device?



> Hi @JamesSmartCell, the test is currently failing and I’m unsure whether it’s due to issues with the test code or if some of the networks are not responding. Could you please verify the network status on the test device?

I removed the ARTIS chain

Now it's giving this error:

```
com.alphawallet.app.RPCNodesTest > should_select_network[Pixel - 10] FAILED 
	androidx.test.espresso.base.DefaultFailureHandler$AssertionFailedWithCauseError: 'not (view is an instance of android.view.ViewGroup and has descendant matching (view.getId() is <2131362276> and has background resource with id 2131230998))' doesn't match the selected view.
	Expected: not (view is an instance of android.view.ViewGroup and has descendant matching (view.getId() is <2131362276/io.stormbird.wallet:id/image_status> and has background resource with id 
```

It might be an idea to wait for 20 seconds or so for all the nodes to return the ping; initially there may not be any status icons visible.

PR #3000, opened 2022-12-11
Hotfix for widget labels overflowing on long strings

justindg:
Hotfix for widget labels overflowing on long strings


justindg:
> @justindg hotfix means it's going out to the app store, right? :)

Might've used the wrong term, editing title

JamesSmartCell (approved):

PR #3002, opened 2022-12-11
Add new OpenSea supported chains.

### LINKED ISSUE #2842: Check out chains supported by OpenSea API and let's
These are supported on the site, let's see if they are supported by the API too and if we need to go to v2 (iOS is still on v1):

* Arbitrum
* Klaytn
* Polygon
* Solana
* Which testnet. Rinkeby or Goerli?

iOS only supports Ethereum and Rinkeby at the moment.

- [x] Check
- [x] Create iOS issue
- [x] Implement Android

JamesSmartCell:
Closes #2842


hboon (approved):
Presume you tested, especially for v2 :)

PR #3014, opened 2022-12-18
Change networkId to hex as per iOS

### LINKED ISSUE #1904: Make chain ID in `engine.chainId` be hex encoded
Refer to https://github.com/AlphaWallet/alpha-wallet-ios/pull/2923/files for implementation.

So it's consistent with:

```
await ethereum.request({method: 'eth_chainId'}) //returns 0x80
```

Note that the value returned by `web3.version.getNetwork()` should still be decimal (empirically determined)

JamesSmartCell:
Closes #1904

### DIFF
--- a/app/src/main/res/raw/init.js
+++ b/app/src/main/res/raw/init.js
@@ -79,8 +79,8 @@ window.AlphaWallet.init(__rpcURL, {
   }
 }, {
     address: __addressHex,
-    networkVersion: __chainID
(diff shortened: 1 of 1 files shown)


hboon (approved):
Sorry, I know why it didn't work the last time now. See https://github.com/AlphaWallet/alpha-wallet-android/issues/1904#issuecomment-1357003017

PR #3019, opened 2022-12-22
Warn Invalid Keystore

### LINKED ISSUE #2694: Should show an error instead of silently importing an invalid Keystore JSON file
Test:

```
{"address":"3f17f1962b36e491b30a40b2405849e597ba5fb5","id":"06bfcaf9-c46e-40b8-ac69-116df63d6eeb","version":3,"crypto":{"cipher":"aes-128-ctr","cipherparams":{"iv":"2efcba77d136cdca01fd8f74f04bb56b"},"ciphertext":"74f8ef051cf13b83d98299e6103edd58f6e7de726568c29306b6ce6a8d81b5eb","kdf":"scrypt","kdfparams":{"dklen":32,"n":4096,"p":6,"r":8,"salt":"daf12812f8c61edaa1897daaa29893e56c2a55177e2c5319a4889a7fb1882aa9"},"mac":"b1067c2fab6fa8203d843b8ac0ee067d995eaa6d9d20020590650cc30682ed91"}}
```

Password:

```
111111
```

(Keystore file above was exported using the wallet that was imported using an invalid key pre-#2689)

Might be directly fixed as part of #2689

JamesSmartCell:
Closes #2694

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/FunctionActivity.java
+++ b/app/src/main/java/com/alphawallet/app/ui/FunctionActivity.java
@@ -174,6 +174,12 @@ public class FunctionActivity extends BaseActivity implements FunctionCallback,
 
         // Fetch attributes local to this action and add them to the injected token properties
         Map<String, TSAction> functions = viewModel.getAssetDefinitionService().getTokenFunctionMap(token.tokenInfo.chainId, token.getAddress());
+        if (functions == null)
+        {
+            recreate();
+            return;
+        }
+
         action = functions.get(actionMethod);
         List<Attribute> localAttrs = (action != null && action.attributes != null) ? new ArrayList<>(action.attributes.values()) : null;
 
--- a/app/src/main/java/com/alphawallet/app/ui/ImportKeystoreFragment.java
+++ b/app/src/main/java/com/alphawallet/app/ui/ImportKeystoreFragment.java
@@ -75,7 +75,10 @@ public class ImportKeystoreFragment extends ImportFragment
     public void onResume()
     {
         super.onResume();
-        if (keystore == null && getActivity() != null) setupView();
+        if (keystore == null || password == null)
+        {
+            requireActivity().recreate();
+        }
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/alphawallet/app/ui/FunctionActivity.java
```diff
@@ -174,6 +174,12 @@ private void getAttrs()
 
         // Fetch attributes local to this action and add them to the injected token properties
         Map<String, TSAction> functions = viewModel.getAssetDefinitionService().getTokenFunctionMap(token.tokenInfo.chainId, token.getAddress());
+        if (functions == null)
```
hboon, line 177: Doesn't seem related to the PR/commit?

hboon: Doe this effectively do a restart/refresh of the activity?

app/src/main/java/com/alphawallet/app/ui/ImportKeystoreFragment.java
```diff
@@ -75,7 +75,10 @@ private void setupView()
     public void onResume()
     {
         super.onResume();
-        if (keystore == null && getActivity() != null) setupView();
+        if (keystore == null || password == null)
+        {
+            requireActivity().recreate();
```
hboon, line 80: Is it possible for this to get into an infinite recursion state?

hboon (approved):

app/src/main/java/com/alphawallet/app/ui/ImportKeystoreFragment.java
JamesSmartCell: > Is it possible for this to get into an infinite recursion state?

This is a signal to the OS to re-create the underlying activity and re-populate the fragment. It happens if the user has paged out from the application while on this page, done something memory intensive like playing a game, watching youtube, navigation etc and paged back in. While they were doing their other activity the OS scavenged the objects in the class, so now they're back, the objects are all null pointers. After the OS re-instantiates the Activity and Fragment the viewodel will have been initialised again.

Really, the Android OS should provide a formal way to deal with this! If they don't already and I don't know about it.

app/src/main/java/com/alphawallet/app/ui/FunctionActivity.java
JamesSmartCell: > Doe this effectively do a restart/refresh of the activity?

Yes, it's an OS directive to do just that. It preserves the starting intents too - so the app will probably just appear to 'flicker' then be in the required state. This ```recreate()``` was found during intensive usage while I was flicking between the app and other apps. The other ```recreate()``` simplifies some other null checking code. This one probably should go into a separate PR.

PR #3047, opened 2023-01-16
Ensure Added Tokens are correctly marked as visible

### LINKED ISSUE #3045: USDC.e disappearing after adding + closing app
Contract: 0xA7D7079b0FEaD91F3e65f86E8915Cb59c1a4C664

JamesSmartCell:
Fixes #3045 

The app has two settings for token visibility:

```enabled``` and ```visibilityChanged```. 
```enabled``` means the app is watching the token but can hide it if it's zero balance. 
```visibilityChanged``` means the user explicitly locked to token as either ```enabled``` = true or ```enabled``` = false.

In this case, the ```visibilityChanged``` and ```enabled``` settings weren't getting updated; it worked correctly for non-zero balances because scanning for the token makes the app aware of that token and it subsequently checks the balance.

TODO: refactor this to use a single enum with:

```
token_disabled,  //never show, don't check token balance
show_token_if_non_zero, //check token balance and show if it becomes positive
always_show_token //show regardless (only if user changes visibility settings)
```


seabornlee (approved):
Great work!

PR #3057, opened 2023-01-23
Check package id when clicking support links

### LINKED ISSUE #3046: Make tapping on support options (URLs) only go to our URLs when it matches our Android app ID
Refer https://github.com/AlphaWallet/alpha-wallet-ios/issues/6143

I think these links: https://github.com/AlphaWallet/alpha-wallet-android/blob/e14255d76621bc26b07c2377352f4c8dcf389db4/app/src/main/java/com/alphawallet/app/entity/MediaLinks.java#L6

justindg:
- Closes #3046

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/MediaLinks.java
+++ b/app/src/main/java/com/alphawallet/app/entity/MediaLinks.java
@@ -1,7 +1,12 @@
 package com.alphawallet.app.entity;
 
+import android.content.Context;
+
 public class MediaLinks
 {
+    // Update these media platform links and ids to target your media groups,
+    // then update the MEDIA_TARGET_APPLICATION to match your applicationId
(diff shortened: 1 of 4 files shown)


JamesSmartCell (changes requested):
In order to make this more whitelabel friendly, could you do a small refactor:

- add our application id to ```class MediaLinks``` at the top; check if the applicationId in MediaLinks matches the current applicationId; if it does then allow the user through to the Social Media platform. Probably move the ```isAlphaWallet``` function in here and change it to ```isMediaTargetted``` or something like that.
- Add a notice in MediaLinks above the ```private String MEDIA_TARGET_APPLICATION="io.stormbird.wallet";``` which reads something like:
```
/* Update these media platform links and ids to target your media groups,
    then update the MEDIA_TARGET_APPLICATION to match your applicationId */
```    
Then we need to add this into the whitelabeling guide.

JamesSmartCell (approved):

PR #3067, opened 2023-02-02
3020 implement token shortcuts

seabornlee:
Long press an NFT in the collection to create up to four shortcuts. If you add a fifth shortcut, the first one will be removed.

- [ ] Show transaction details
- [ ] Show sender wallet

### DIFF
--- a/app/src/main/java/com/alphawallet/app/C.java
+++ b/app/src/main/java/com/alphawallet/app/C.java
@@ -235,8 +235,8 @@ public abstract class C {
     public static final String LINKEDIN_PACKAGE_NAME = "com.linkedin.android";
     public static final String REDDIT_PACKAGE_NAME = "com.reddit.frontpage";
     public static final String INSTAGRAM_PACKAGE_NAME = "com.instagram.android";
-
     public interface ErrorCode {
+
         int UNKNOWN = 1;
         int CANT_GET_STORE_PASSWORD = 2;
         int ALREADY_ADDED = 3;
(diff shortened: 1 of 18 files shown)


seabornlee:
@JamesSmartCell Does it make sense to transfer a NFT which opened by shortcut and belongs to another wallet, and pay the gas with the currently active wallet?

JamesSmartCell:
> @JamesSmartCell Does it make sense to transfer a NFT which opened by shortcut and belongs to another wallet, and pay the gas with the currently active wallet?

That sounds terrifying! I think people would just want to change to that wallet and jump to the token. If the shortcut was to a high value NFT and I clicked on the shortcut and it transferred the NFT I'd be quite surprised (and if I didn't know it had done that I would think I had lost my NFT).

JamesSmartCell:
@seabornlee if I have two wallets, Wallet A and Wallet B. The token I use to operate my front door is in Wallet A, but I use mostly Wallet B for testing. When I use the TokenScript for the front door, after using the shortcut I click on the function button in the screen that takes me to FunctionActivity. This then tries to fetch the token passed in, but because the default wallet is different from the wallet that was called from the shortcut the token will be null, since it's fetching a specific tokenId and it's trying to fetch it from Wallet B (since that's the default wallet).

- (easiest and most reliable) change the default wallet when the shortcut is pressed. Caching and then reverting back to the initial wallet when the NFTAssetDetailActivity is destroyed should work and should have no consequences.
- pass in the working wallet to all NFT leaf Activities and use that. This isn't a bad solution really and is more deterministic.

JamesSmartCell:
@seabornlee to get this to merge we need to have all the leaf activities using the wallet passed via the activity intent - I think only FunctionActivity is outstanding.

seabornlee:
Open a new PR #3181  instead.

PR #3072, opened 2023-02-06
revise nft ui

seabornlee:
image


JamesSmartCell (changes requested):
In general it looks like a good improvement!

Just a few teething issues:

1. Token thumbnails should be the same size, rather than changing to accommodate the aspect ratio.
2. Set the thumbnail size before populating the image so no matter how long it takes there's no crunch and expanding of tokens.
3. Some tokens don't have the name bar populated - it's usually the first token in the list if that helps.

JamesSmartCell (approved):

PR #3075, opened 2023-02-08
OKX Network Integration

### LINKED ISSUE #3065: Add Okx chain
Fetch the relevant RPC data from chainlists.org

Network Name: OKXChain Mainnet
RPC:https://exchainrpc.okex.org/
Chain ID: 66
Currency Symbol: OKT 
Block Explorer URL: https://www.oklink.com/en/okc

I am getting an API key from the team, as soon as I have it you can add Etherscan Transaction fetch. 

It appears they don't have a raw transaction fetch (etherscanAPI) however they do have the much more useful transfer fetch. This is the one that is fetched here:

```public Single<Integer> readTransfers(String walletAddress, NetworkInfo networkInfo, TokensService svs, boolean isNFTCheck)```

You would need to write a custom fetch similar to ```processCovalentEvents``` where covalent events are converted into etherscan events then the ```private int processEtherscanEvents(``` function is called on the converted events.

This will allow us to find the NFTs and ERC20 tokens; the wallet will populate those using the RPC. The wallet will also be able to fetch the corresponding transaction from RPC and add that directly.

Here's the relevant API page:

https://www.oklink.com/docs/en/#on-chain-data-api-address-query-address-transaction-list-information

Soo as I get the API key I'll forward it to you. Let me know if you have any issues...

### PR DESCRIPTION
**Pull request by justindg**
- Closes #3065 

First draft of OKX Integration

### CODE REVIEW

File: app/src/main/java/com/alphawallet/app/entity/NetworkInfo.java [hunk 1]
```diff
@@ -55,7 +56,7 @@ public String getShortName()
     public boolean usesSeparateNFTTransferQuery()
     {
         return (etherscanAPI != null && !etherscanAPI.contains(BLOCKSCOUT_API)
-                && !etherscanAPI.contains(COVALENT) && !etherscanAPI.contains(PALM_API));
+                && !etherscanAPI.contains(COVALENT) && !etherscanAPI.contains(PALM_API) && chainId == OKX_ID);
     }
```
JamesSmartCell, line 61: this logic breaks all the other chains - should just not put ```chainId == okx``` it'll return true

JamesSmartCell (changes requested):
Hi @justindg, can you check the ```//JB:``` comments I added:

- maybe make OKLinkService into a static class with static methods.
- in TokenRepository return TokenInfo from OKX to save RPC calls.
- fold API calls

This should improve the pickup of Tokens by quite a bit.

JamesSmartCell (approved):

PR #3079, opened 2023-02-10
Fix Event Sync

### LINKED ISSUE #3078: Event Sync fail
On certain chains (Polygon esp) the event sync is not working correctly.

Further, the limit discovery system appears to be blocking.

JamesSmartCell:
Fixes an issue where esp Polygon event sync failed on two counts:
- Batch event fetch fails (add system where we dynamically track and cache failure to fetch batch array format).
- Sync size wasn't being allowed to dynamically change if an incorrect limit is applied.

Fixes #3078

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/EventSync.java
+++ b/app/src/main/java/com/alphawallet/app/entity/EventSync.java
@@ -35,14 +35,14 @@ import timber.log.Timber;
 public class EventSync
 {
     public static final long BLOCK_SEARCH_INTERVAL = 100000L;
-    public static final long POLYGON_BLOCK_SEARCH_INTERVAL = 10000L;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/alphawallet/app/entity/EventSync.java
```diff
@@ -174,7 +174,7 @@ private BigInteger adjustForLimitedBlockSize(BigInteger eventReadStartBlock, Big
 
     public boolean handleEthLogError(Response.Error error, DefaultBlockParameter startBlock, DefaultBlockParameter endBlock, SyncDef sync, Realm realm)
     {
-        if (error.getCode() == -32005)
+        if (error.getCode() == -32005 || error.getCode() == -32600)
```
seabornlee, line 177: I suggest extracting constants or adding comments to express the errors they are dealing with.

seabornlee (approved):

PR #3086, opened 2023-02-13
Pick up ERC1155 from transfer APIs

### LINKED ISSUE #3062: Check ERC1155 detection
https://discord.com/channels/548357107267928064/1067548127504113765/1069454731317084260

Just in case it's lost

JamesSmartCell:
Refactor the way ERC1155 tokens are picked up. This also required a little re-work of some other areas to ensure it works correctly.

Etherscan and other TX APIs added a new ```token1155tx``` action so the simple boolean NFt toggle needed to be removed and handled a different way.

Fixes #3062

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/transactionAPI/TransferFetchType.java
+++ b/app/src/main/java/com/alphawallet/app/entity/transactionAPI/TransferFetchType.java
@@ -1,0 +1,11 @@
+package com.alphawallet.app.entity.transactionAPI;
+
+/**
+ * Created by JB on 10/02/2023.
+ */
+public enum TransferFetchType
+{
+    tokentx,
+    tokennfttx,
+    token1155tx
+}
--- a/app/src/main/java/com/alphawallet/app/entity/EtherscanEvent.java
+++ b/app/src/main/java/com/alphawallet/app/entity/EtherscanEvent.java
@@ -1,11 +1,14 @@
 package com.alphawallet.app.entity;
 
+import android.text.TextUtils;
+
 import com.alphawallet.app.repository.TokenRepository;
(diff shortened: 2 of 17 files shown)


app/src/main/java/com/alphawallet/app/entity/transactionAPI/TransferFetchType.java
```diff
@@ -0,0 +1,11 @@
+package com.alphawallet.app.entity.transactionAPI;
+
+/**
+ * Created by JB on 10/02/2023.
+ */
+public enum TransferFetchType
```
justindg, line 6: Not a big deal but it would be nice to have enums follow a more standard/flexible format

See `ImportWalletType` class

JamesSmartCell: Yeah that's good - I didn't know you could do that!

JamesSmartCell: Please feel free to refactor when you build on this for OKX.

seabornlee (approved):
Later maybe we can refactor to have subclass for Token such as ERC1155Token.

seabornlee:
> > Later maybe we can refactor to have subclass for Token such as ERC1155Token.
> 
> Do you mean move the code to the ERC1155Token class or is this in a different context, a different Token class for use only in the Transaction fetch?
> 
> The UpdateTokenBalance method does need that refactor.

Oh, we already have an ERC1155Token subclass, but it's large enough. I'm considering having a strategy class for each type of token. To gain a better understanding, we need to read the code in more detail.

PR #3096, opened 2023-02-17
Optimise Transaction API usage

JamesSmartCell:
- Optimise OKX link operation so the API isn't hammered as much (and less network activity/load from AW).
- Improve the code in TransactionsNetworkClient

### DIFF
--- a/app/src/main/java/com/alphawallet/app/service/OkLinkService.java
+++ b/app/src/main/java/com/alphawallet/app/service/OkLinkService.java
@@ -81,30 +81,36 @@ public class OkLinkService
         return JsonUtils.EMPTY_RESULT;
     }
 
-    public EtherscanEvent[] getEtherscanEvents(String address, TransferFetchType tfType)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/alphawallet/app/service/OkLinkService.java
```diff
@@ -123,6 +129,21 @@ public EtherscanEvent[] getEtherscanEvents(String address, TransferFetchType tfT
         return etherscanEvents.toArray(new EtherscanEvent[0]);
     }
 
+    //@SuppressWarnings("ConstantConditions")
+    private boolean compareEventsWithLastRead(List<OkxEvent> events, long lastBlockRead)
+    {
+        for (OkxEvent ev : events)
+        {
+            long height = Long.parseLong(ev.height);
```
justindg, line 137: Might be a good idea to check if `ev.height` is null or empty (via `TextUtils.isEmpty()`) to avoid a potential crash

justindg (approved):

PR #3105, opened 2023-02-21
Optimise & simplify token display updates

JamesSmartCell:
This PR works together with the recent fix for the TokenID collision issue.

Now, only tokens that have had their balance changed are updated.

Bonus is that it removes a Realm listener which was being flagged by the profiler. Shame because in theory the Realm listener approach is the most elegant solution.

### DIFF
--- a/app/src/main/java/com/alphawallet/app/repository/TokensRealmSource.java
+++ b/app/src/main/java/com/alphawallet/app/repository/TokensRealmSource.java
@@ -323,7 +323,6 @@ public class TokensRealmSource implements TokenLocalSource {
     private void setTokenUpdateTime(RealmToken realmToken, Token token)
     {
         realmToken.setLastTxTime(System.currentTimeMillis());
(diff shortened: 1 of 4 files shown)


seabornlee:
When I swapped all my ENS for stETH, my ENS balance was set to zero, but it didn't automatically remove itself from the list.

seabornlee (approved):

JamesSmartCell:
> I found a new issue, when I swapped all my ENS for stETH, my ENS balance was set to zero, but it didn't automatically remove itself from the list. It probably related to code:
> 
> ```
> if (meta.balance.equals("-2"))
> {
>     metasToDelete.add(meta);
> }
> ```

This should only be set if the code detects the token has been deleted. Previously there was a bug that if you had bad network it could set the balance to -2 which has been fixed now.

If the ENS balance is still showing it might be because the user visibility flag (```visibilityChanged```) has been set. It would be useful to have a visual indication in debug mode if this flag is set. At least in the TokenManagement page.

PR #3107, opened 2023-02-21
Fix "Connection timeout..." when starting a new WC session

### LINKED ISSUE #3106: WalletConnect "Connection attempt timed out..." when trying to start a session
**Site**: 
https://example.walletconnect.org/

When attempting to start a new session, "Connection attempt timed out..." dialog appears.  
This particular dialog is shown when `WCClient.chainIdVal()` returns `0`.

**Diagnosis**:  
The payload from this particular session request has `"chainId":null` , and when this is the case, `WCClient` defaults `chainId` to `0`.

Interim solution is to default to mainnet ID, when null or empty string is detected.

justindg:
- Closes #3106 

Default to mainnet ID instead of zero value


JamesSmartCell:
Nice, yeah it's ok to assume mainnet to get it going - I found a lot of WC implementations are not written very well - no chain is specified.

WC1 has been deprecated already but it seems hardly anyone has switched to WC2 yet.

JamesSmartCell (approved):

Issue #3104, opened 2023-02-21
Create test case that fails with the message to revisit issue about localizing decimal and thousands separator


hboon:
```
fun testRevisitIssue()
    let now = Date()
    if now >= "20240221"
        fail("Revisit https://github.com/AlphaWallet/alpha-wallet-android/issues/3099#issuecomment-1437751379. Create issue and remove this test")
    end
end
```

Yes, in the test suite :)

PR #3108, opened 2023-02-21
Goerli etherscan doesn't have ERC1155 route

JamesSmartCell:
Goerli Etherscan doesn't have the ERC1155 tx route. 
This prevents that route getting needlessly called.
We should check from time to time if it starts working.


hboon:
> check from time to time

@JamesSmartCell can we abuse the test suite for this too? Like https://github.com/AlphaWallet/alpha-wallet-android/issues/3104. It’s the only way that survives any bus factor

seabornlee (approved):

PR #3119, opened 2023-02-24
WalletConnect Improvements

### LINKED ISSUE #3090: Check that chain ID in EIP-712 message is the same as the current/enabled chain before showing user the prompt to sign it
- [ ] A. in the dapp browser and if not using WalletConnect (if the chain ID is different from the current one in the browser), return or display an error to the dapp. Display an error message to the user, something like: "You were requested to sign a message for the Ethereum [replace with the target chain name] chain. You are currently on the Polygon [replace with current chain name] chain"
- [ ] B. for WalletConnect v1 (if the chain ID is different from the current one for the WC session), return error to WalletConnect server. Display an error message to the user, something like: "You were requested to sign a message for the Ethereum [replace with the target chain name] chain. You are currently on the Polygon [replace with current chain name] chain"
- [ ] C. for WalletConnect v2 (if the chain ID is not included in the list of active chains for the WC session), return error to WalletConnect server. Display an error message to the user, something like: "You were requested to sign a message for the Ethereum [replace with the target chain name] chain. You need to enable this chain for your WalletConnect session"

### LINKED ISSUE #3111: EthereumNetworkBase.getShortChainName(long chainId) fallback value is MAINNET_ID (Ethereum)
If for some reason an unsupported/removed network id is supplied, i.e. `42` (Kovan), the fallback value is is MAINNET_ID ("Ethereum")

A more reasonable return value should be an empty string, and the method caller should handle it.

--

Related issue: #3090

In this issue, there is a chance where we pull the the network name via an unsupported network ID, therefore it must not return "Ethereum"

### LINKED ISSUE #3116: Ethereum is the default network in ActionSheet when connecting via WalletConnect, despite being disabled

### LINKED ISSUE #3118: Display error dialog when WalletConnect Session Request is from an unsupported network
There could be edge cases where the WC session request is from an unsupported chain, or a previously deprecated chain (i.e. Kovan, Rinkeby, etc)

Error message: `Unable to handle session request from an unsupported chain (id: <CHAIN_ID>).`

justindg:
- Closes #3090 
- Closes #3118 
- Closes #3116 
- Closes #3111

---

### DIFF
--- a/app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
+++ b/app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
@@ -1199,7 +1199,7 @@ public class DappBrowserFragment extends BaseFragment implements OnSignTransacti
                 getString(R.string.error_eip712_incompatible_network,
                     EthereumNetworkBase.getShortChainName(requestingChainId),
                     EthereumNetworkBase.getShortChainName(activeChainId)) :
-                getString(R.string.error_eip712_unsupported_network);
+                getString(R.string.error_eip712_unsupported_network, String.valueOf(requestingChainId));
             errorDialog.setMessage(message);
             errorDialog.setButton(R.string.action_cancel, v -> {
                 errorDialog.dismiss();
--- a/app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
+++ b/app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
@@ -55,6 +55,7 @@ import androidx.annotation.Nullable;
 import androidx.appcompat.widget.Toolbar;
 import androidx.core.content.ContextCompat;
 import androidx.fragment.app.Fragment;
+import androidx.lifecycle.Lifecycle;
 import androidx.lifecycle.ViewModelProvider;
 
 import com.alphawallet.app.C;
@@ -76,6 +77,7 @@ import com.alphawallet.app.entity.WalletType;
 import com.alphawallet.app.entity.analytics.ActionSheetSource;
 import com.alphawallet.app.entity.analytics.QrScanResultType;
 import com.alphawallet.app.entity.tokens.Token;
+import com.alphawallet.app.repository.EthereumNetworkBase;
 import com.alphawallet.app.repository.EthereumNetworkRepository;
 import com.alphawallet.app.repository.TokenRepository;
 import com.alphawallet.app.repository.TokensRealmSource;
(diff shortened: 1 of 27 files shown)


app/src/main/java/com/alphawallet/app/ui/DappBrowserFragment.java
```diff
@@ -1176,13 +1179,39 @@ private void showChainChangeDialog(long callbackId, NetworkInfo newNetwork)
 
     private void handleSignMessage(Signable message)
     {
-        if (confirmationDialog == null || !confirmationDialog.isShowing())
+        if (message.getMessageType() == SignMessageType.SIGN_TYPED_DATA_V3 && message.getChainId() != activeNetwork.chainId)
+        {
```
JamesSmartCell, line 1183: Does V4 have the chainId? As far as I'm aware the only difference between V3 and V4 is that V4 supports arrays, so we should check for SIGN TYPE DATA rather than SIGN TYPE DATA V3. Unless V3 also covers V4.

app/src/main/java/com/alphawallet/app/ui/WalletConnectActivity.java
```diff
@@ -791,14 +790,27 @@ private void onSessionRequest(Long id, WCPeerMeta peer, long chainId)
...
+            confirmationDialog.fullExpand();
+
+            viewModel.track(Analytics.Action.WALLET_CONNECT_SESSION_REQUEST);
+
+            if (!viewModel.isActiveNetwork(chainId) && !viewModel.isActiveNetwork(chainIdOverride))
+            {
+                openChainSelection();
+            }
+        }
     }
```
JamesSmartCell, line 811: Can this method be refactored to remove the nested conditionals? Maybe split it up a bit.

app/src/main/java/com/alphawallet/app/walletconnect/WalletConnectV2SessionRequestHandler.java
```diff
@@ -61,19 +68,74 @@ private void showDialog(String method, ActionSheetCallback aCallback)
...
-            Signable signable = signRequest.getSignable(sessionRequest.getRequest().getId(), settledSession.getMetaData().getUrl());
-            ActionSheet actionSheet = new ActionSheetSignDialog(activity, aCallback, signable);
-            actionSheet.setSigningWallet(signRequest.getWalletAddress());
-            List<String> icons = Objects.requireNonNull(settledSession.getMetaData()).getIcons();
-            if (!icons.isEmpty())
+            Signable signable = signRequest.getSignable(sessionRequest.getRequest().getId(), Objects.requireNonNull(settledSession.getMetaData()).getUrl());
+            WalletConnectV2SessionItem session = new WalletConnectV2SessionItem(settledSession);
+
+            List<Long> chainList = new ArrayList<>();
+            for (String chain : session.chains)
```
JamesSmartCell, line 75: Refactor to make this a bit more maintainable. Maybe move the chainList formation into a separate method.

JamesSmartCell (changes requested):
This looks like it'll fix quite a few outstanding issues.

Could you tidy up the code a little before we commit it?

JamesSmartCell (approved):

PR #3190, opened 2023-04-11
Reload HomeActivity when active wallet changed

### LINKED ISSUE #3161: The defaultWallet in WalletViewModel is not actively updated after account deletion
Need to manually pull down to update，If I don't manually update, click on the token to jump to Erc20DetailActivity to display the data of the deleted wallet

### LINKED ISSUE #3182: Delete wallet causes crash
**Steps:**
Create a new wallet
Transfer a NFT to the new wallet
View the NFT in new wallet
Delete the new wallet
Go to homepage

**Expected:**
The list is updated to the current active wallet

**Actual:**
It is the list of the deleted wallet, and clicking the NFT collection results in a crash

seabornlee:
Close #3161 
Close #3182

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/Wallet.java
+++ b/app/src/main/java/com/alphawallet/app/entity/Wallet.java
@@ -9,6 +9,7 @@ import com.alphawallet.app.service.KeyService;
 import com.alphawallet.app.util.BalanceUtils;
 
 import java.math.BigDecimal;
+import java.util.Objects;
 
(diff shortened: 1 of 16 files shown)


JamesSmartCell (changes requested):
Small issue (not a showstopper, but it's noticable):

1. Delete active wallet (should now be in the "Wallets" page, with the first Seed phrase wallet selected).
2. Select a different wallet.
3. Return to home page - it now resets twice.

PR #3208, opened 2023-04-20
Import and use Attestations

JamesSmartCell:
Stage 1 of using attestations as tokens.

Still needs a little tidying up but this stage needs to go in before it attracts more merge conflicts.


app/src/main/java/com/alphawallet/app/entity/tokens/Attestation.java
```diff
@@ -0,0 +1,131 @@
...
+        issuerKey = attValidation._issuerKey;
+    }
+
+    public AttestationValidationStatus isValid()
+    {
+        if (!isValid)
+        {
+            return AttestationValidationStatus.Expired;
+        }
+        else if (!TextUtils.isEmpty(attestationSubject) && !attestationSubject.equalsIgnoreCase(getWallet()))
```
seabornlee, line 72: These `else` are redundant.

app/src/main/java/com/alphawallet/app/service/AssetDefinitionService.java
```diff
@@ -632,6 +650,41 @@ public Single<Boolean> refreshAttributes(Token token, TokenDefinition td, BigInt
...
+                                functionResults.deleteAllFromRealm();
+                            });
+                        }
+                    }
+
+                }
+            }
+            catch (Exception e)
+            {
+                e.printStackTrace();
```
seabornlee, line 681: Timber.e()

seabornlee (approved):

app/src/main/java/com/alphawallet/app/entity/tokens/Attestation.java
JamesSmartCell: Hi @seabornlee could you elaborate? This function returns a validation status which could either be "Pass" if everything checks out, or it could have failed in one of 3 ways:

- The attestation has expired (outside of the time gate).
- The attestation subject (who it was issued to) does not match the importing wallet.
- The issuer of the attestation is not correct (it was signed by an unrecognised key).

PR #3227, opened 2023-05-09
Display transaction failure error message

### LINKED ISSUE #3209: Add the reason for transaction failure to the "fail notification" dialog popup
Like Remix does. Usually contract authors will include a failure message if a 'require' in the Solidity fails, if this message seen we should show it to the user as the reason why the transaction will fail.

Need to modify the estimate function, probably make this:

```
public Single<BigInteger> calculateGasEstimate(byte[] transactionBytes, long chainId, String toAddress,
                                                   BigInteger amount, Wallet wallet, final BigInteger defaultLimit)
```

into ```Single<Pair<BigInteger, String>> calculateGasEstimate(byte[] transactionBytes, long chainId, String toAddress,
                                                   BigInteger amount, Wallet wallet, final BigInteger defaultLimit)```

Where the String will return the error message from the ```EthEstimateGas``` object, ```estimate.getError().getMessage()```. 

Propagate that back down though to all the dialog warnings, and display the error on those warning dialog boxes instead of the generic 'transaction will fail' popup.

justindg:
- Closes #3209

### DIFF
--- a/app/src/main/java/com/alphawallet/app/entity/ActionSheetInterface.java
+++ b/app/src/main/java/com/alphawallet/app/entity/ActionSheetInterface.java
@@ -4,8 +4,6 @@ import androidx.activity.result.ActivityResult;
 
 import com.alphawallet.app.web3.entity.Web3Transaction;
 
-import java.math.BigInteger;
-
 /**
  * Created by JB on 16/01/2021.
  */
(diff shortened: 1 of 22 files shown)


JamesSmartCell:
Looks great @justindg. I'll test it later today along with the notification PR.

JamesSmartCell (changes requested):
The dialog should be in this format:

### Gas estimation failed

Gas estimation errored: "execution reverted: ERC721: transfer caller is not owner nor approved"
The transaction execution will likely fail. Do you want to force sending?

Then restore the previous two buttons, either "Proceed" or "Cancel". Where proceed lets the user still try to push the tx as before.

JamesSmartCell (approved):

PR #3234, opened 2023-05-17
3149 Handle namespace with WC utility function

seabornlee:
https://docs.walletconnect.com/2.0/android/web3wallet/wallet-usage#namespace-utils
Tested with: https://lab.web3modal.com/ManagedReact


JamesSmartCell (commented):
Hi @seabornlee this works perfectly on my end. 

One question (probably unrelated to this PR) when I click on the active WalletConnect entry in the wallet fragment, and a walletconnect 2 session is active - should it take me to the WalletConnectV2Activity screen, but in an information context? Currently it goes to the old legacy walletconnect screen and my active session isn't showing there.

JamesSmartCell (approved):

